PR #4, opened 2020-12-09
Support adding units

norhh:


src/spec-lang/ast/number.ts
```diff
@@ -1,14 +1,35 @@
...
+    ["minutes", bigInt(60)],
+    ["hours", bigInt(3600)],
+    ["days", bigInt(86400)],
+    ["weeks", bigInt(604800)]
+]);
 
 export class SNumber extends SNode {
     public readonly num: BigInteger;
     public readonly radix: number;
+    private readonly conversionMultipler: Map<string, BigInteger> = units;
```
cd1m0, line 19: This private variable is unnecessary. Should just rename the file-level `units` constant to `unitMultiplier` and use that instead.

cd1m0 (changes requested):
Good work! Just one nit (will fix it and push)

src/spec-lang/ast/number.ts
cd1m0: Fixed in 4f5f56a

cd1m0 (dismissed):

cd1m0 (approved):

PR #5, opened 2020-12-10
Disallow if_succeeds annotation for contract definitions

blitz-1306:
## Changes
- [x] Disallowed any annotation types, except `invariant` type on contracts.
- [x] Disallowed any annotation types, except `if_succeeds` type on functions.
- [x] Overhauled `AnnotationExtractor`:
    - Isolated annotation parsing to `makeAnnotationFromMatch()` funtion for a better readability.
    - Introduced `validateAnnotation()` method to check that created annotation is valid for the target. Moved rules from scribble executable module to this method to concentrate logic in a single place.
    - Tweaked `findAnnotations()` to respect filtering in-place and do not trigger validation on filtered out annotations.
- [x] Introduced `AnnotationError` class. The `SyntaxError` is now based on `AnnotationError`. Also introduced `UnsupportedByTargetError` class for validation errors.
- [x] Minor cleanup in scribble executable module to adapt to a new logic.
- [x] Fixed affected test samples.
- [x] Introduced test cases to check that errors are actually thrown.

Regards.

### DIFF
--- a/src/bin/scribble.ts
+++ b/src/bin/scribble.ts
@@ -264,6 +264,9 @@ function instrumentFiles(
         assert(contents !== undefined, `Missing source for ${unit.absolutePath}`);
 
         for (const fun of unit.vFunctions) {
+            /**
+             * We call `getAnnotationsOrDie()` here to make sure there are no annotations on free functions.
+             */
             getAnnotationsOrDie(fun, ctx.files, filters);
         }
 
--- a/src/bin/scribble.ts
+++ b/src/bin/scribble.ts
@@ -206,6 +206,12 @@ function computeContractInvs(
                 error(`Unsupported contract annotations on ${contract.kind} ${contract.name}`);
             }
 
+            if (annotations.some((annotation) => annotation.type === AnnotationType.IfSucceeds)) {
+                error(
+                    `Annotation type "${AnnotationType.IfSucceeds}" is not applicable to contracts (specified for ${contract.kind} ${contract.name})`
+                );
+            }
+
             contractAnnotMap.set(contract, annotations);
         }
     });
--- a/src/bin/scribble.ts
+++ b/src/bin/scribble.ts
@@ -1,4 +1,6 @@
 #!/usr/bin/env node
+import fse from "fs-extra";
+import path, { dirname, relative } from "path";
 import {
     ASTContext,
     ASTNodeFactory,
@@ -21,14 +23,12 @@ import {
     UserDefinedTypeName,
     VariableDeclaration
 } from "solc-typed-ast";
-import fse from "fs-extra";
-import path, { dirname, relative } from "path";
 import { print, rewriteImports } from "../ast_to_source_printer";
 import {
     Annotation,
     AnnotationExtractor,
-    AnnotationType,
-    SyntaxError
+    SyntaxError,
+    UnsupportedByTargetError
 } from "../instrumenter/annotations";
 import { getCallGraph } from "../instrumenter/callgraph";
 import { CHA, chaDFS, getCHA } from "../instrumenter/cha";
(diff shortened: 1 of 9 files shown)


blitz-1306:
@cd1m0 So, should we extend the implementation to check invariants over functions? It seems that it is related to https://github.com/ConsenSys/scribble/issues/6#issuecomment-744537926.

cd1m0:
Yes! We should definitely extend the check to make sure that you can't specify an `invariant` annotation on a function

cd1m0 (approved):
The changes so far look good. Just need to add a similar check that `invariant` doesn't show up on function annotations.

blitz-1306:
@cd1m0 I bit reworked the `AnnotationExtractor` in attemp to concentrate validation rules in one place. Also the changes (in my opinion) will make flow a bit more clear. Here is the simplified flow for annotation extraction:
- Find match in structured documentation.
- Proceed with annotation creation (delimiter check, syntax check and parse predicate expression).
- Check if annotation is passing filter rules. If it not passing, then skip further steps.
- Validate annotation.
- Push annotation to the result array.

We can also consider as an option to filtering prior to annotation parsing to not do additional work for the filtered out annotations. However for now it may be better to do syntax checking at least. So I leave the decision up to you on this.

Regards.

src/bin/scribble.ts
```diff
@@ -267,25 +264,14 @@ function instrumentFiles(
...
-
-                prettyError(
-                    "UnsupportedError",
-                    "Instrumenting free functions is not supported yet",
-                    unit,
-                    annotation.predicateFileLoc(contents),
-                    annotation.original
-                );
-            }
+            getAnnotationsOrDie(fun, ctx.files, filters);
```
cd1m0, line 267: Should add a comment here to explain the logic:

// We call getAnnotationsOrDie to make sure there are no annotations on free functions

src/instrumenter/annotations.ts
```diff
@@ -137,161 +149,210 @@ export class SyntaxError extends Error {
...
+        }
+    }
 
-                    throw new SyntaxError(e.message, original, errRange);
-                }
+    private validateAnnotation(target: AnnotationTarget, annotation: Annotation, source: string) {
+        if (target instanceof ContractDefinition) {
+            if (annotation.type !== AnnotationType.Invariant) {
+                throw new UnsupportedByTargetError(
+                    `Annotation type "${annotation.type}" is not applicable to a target`,
```
cd1m0, line 276: Small rewording:

`"${annotation.type}" annotation is not applicable to contracts`

src/instrumenter/annotations.ts
```diff
@@ -137,161 +149,210 @@ export class SyntaxError extends Error {
...
+                throw new UnsupportedByTargetError(
+                    `Unsupported contract annotations on ${target.kind} ${target.name}`,
+                    annotation.original,
+                    annotation.annotationFileLoc(source)
+                );
+            }
+        } else if (target instanceof FunctionDefinition) {
+            if (annotation.type !== AnnotationType.IfSucceeds) {
+                throw new UnsupportedByTargetError(
+                    `Annotation type "${annotation.type}" is not applicable to a target`,
```
cd1m0, line 292: Small rewording:

"${annotation.type}" annotation is not applicable to functions

cd1m0 (approved):
Awesome work! 2-3 small re-wordings. Will fix them right now.

src/instrumenter/annotations.ts
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/5/commits/3b1a65c0b0e464efe1aaada050bfe36bd5592678.

src/instrumenter/annotations.ts
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/5/commits/3b1a65c0b0e464efe1aaada050bfe36bd5592678.

src/bin/scribble.ts
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/5/commits/3b1a65c0b0e464efe1aaada050bfe36bd5592678.

cd1m0 (approved):
LGTM

PR #8, opened 2021-01-05
0.3.3 

cd1m0:
Small patch release. Includes:

 - disallow `if_succeeds` from contract annotations and `invariant` from function annotations
 - fix for buggy renaming of identifier paths in user defined type names in flat mode
 - nits

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -9,10 +9,11 @@ steps:
     - name: build_test
       image: node:10
       environment:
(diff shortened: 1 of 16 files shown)


cd1m0:
Drone CI failed due to tests  taking too long. I saw the run succesfully so force-pushing (in a bit of a hurry to release this fix). We should do something to speed-up tests.

PR #10, opened 2021-01-17
FS 25

norhh:
TODO: -

- [ ] Add tests.(Will do it after confirming that I am not doing something which is completely unintended)


norhh:
Nevermind, there was a problem

PR #11, opened 2021-01-19
FS 25

norhh:


src/instrumenter/instrument.ts
```diff
@@ -978,15 +1017,31 @@ function insertVarsStruct(
...
+    const allNames = getAllNames(contract);
+    const funcName = `__scribble_${contract.name}_check_state_invariants_internal`;
+    if (!allNames.has(funcName)) {
+        return funcName;
+    }
+
+    let idx = 0;
+    while (allNames.has(`${funcName}_${idx}`)) {
+        idx++;
+    }
```
blitz-1306, line 1042: I see that we have several occurances of the same logic within different functions. While I do understand the reasoning and approach, I also believe that we can dedicated utility function to reduce code duplication. WDYT?

src/instrumenter/instrument.ts
```diff
@@ -530,7 +557,14 @@ export function generateExpressions(
     // Step 1: Define struct holding all the temporary variables neccessary
     const exprs = annotations.map((annot) => annot.expression);
     const factory = ctx.factory;
-    const structName = uid.get("vars");
+    const allNames = getAllNames(contract);
+
+    let possibleStructName = uid.get("vars");
+    while (allNames.has(possibleStructName)) {
```
cd1m0, line 563: We can use `getNewVarName(possibleStructNames, "vars")` here instead of the while loop.

src/instrumenter/instrument.ts
```diff
@@ -127,6 +134,26 @@ const CHECK_STATE_INVS_FUN = "__scribble_check_state_invariants";
...
 
+function getAllNames(contract: ContractDefinition): Set<string> {
+    const nameSet: Set<string> = new Set();
+    for (const v of contract.getChildren()) {
+        if ("name" in v) nameSet.add(v["name"]);
+    }
+    return nameSet;
+}
+
+function getNamesInFuncScope(contract: ContractDefinition, fn: FunctionDefinition): Set<string> {
```
cd1m0, line 145: There are several additional categories of declared names, that are missing here:

 - structs, events and enums declared in the current contract
 - structs, enums and constants declared at the source unit level
 - names that are imported (and potentially renamed) from other files.
 
Getting the list of all declared names in a function scope, (or any scope) is thus a little tricky :(. We have some code that tries to compute this in `maru-ir`. The more I think about it, perhaps its better we relocate that code to `solc-typed-ast`, and rebase `scribble` to use it.

src/instrumenter/instrument.ts
```diff
@@ -127,6 +134,26 @@ const CHECK_STATE_INVS_FUN = "__scribble_check_state_invariants";
 const OUT_OF_CONTRACT = "__scribble_out_of_contract";
 const CHECK_INVS_AT_END = "__scribble_check_invs_at_end";
 
+function getAllNames(contract: ContractDefinition): Set<string> {
```
cd1m0, line 137: I think this function might not include names implicitly included as part of an import. E.g. in the below example the enum definition `vars0` from `lib.sol` is implicitly shadowed in `foo.sol` which imports `lib.sol`.

lib.sol

enum vars0 {
  A,B,C
}

foo.sol:
import "lib.sol"
contract Foo {
   struct vars0 {...};
   function foo() {
      vars0 _v;
      ...
   }
}

src/instrumenter/instrument.ts
```diff
@@ -978,15 +1017,35 @@ function insertVarsStruct(
...
+    contractInvariantsData: ContractInvariantsData | undefined
+) {
+    if (contractInvariantsData) {
+        return contractInvariantsData.internalInvariantFunction;
+    }
+
+    const allNames = getAllNames(contract);
+    let funcName = `__scribble_${contract.name}_check_state_invariants_internal`;
+
+    if (!allNames.has(funcName)) {
```
cd1m0, line 1043: This pattern appears in several places:

`if (!someNames.has(name)) {
   return name;
 }
 
 name = getNewVarName(names, "${name}_");
 `
 
 Perhaps we can roll that `if` inside of `getNewVarName`, such that `getNewVarName(names, str)` returns `str` when its not inside `names`, and otherwise returns `${str}_[0-9]*`

src/instrumenter/instrument.ts
```diff
@@ -978,15 +1017,35 @@ function insertVarsStruct(
 }
 
 function getCheckStateInvsFuncs(contract: ContractDefinition): FunctionDefinition {
-    return single(contract.vFunctions.filter((fn) => fn.name === CHECK_STATE_INVS_FUN));
```
cd1m0, line 981: This function is a bit fragile. Instead of relying on reg-exes to find the the check invariants functions, and duplicating the knowledge of the exact shape of those functions, lets remove this function all-toghether, and instrumentedInvariantData as an argument to `FunctionInstrumenter.instrument`. This is the only place where this function is used.

src/instrumenter/interpose.ts
```diff
@@ -82,16 +82,34 @@ function callOriginal(
...
+function renameReturns(
+    factory: ASTNodeFactory,
+    stub: FunctionDefinition,
+    varsInScope: Set<string>
+): Recipe {
+    const returnVars = stub.vReturnParameters.vParameters.filter((ret) => ret.name === "");
+    let varId = 0;
+    const renamedVars = [];
+    for (let i = 0; i < returnVars.length; i++) {
+        while (varsInScope.has(`RET_${varId}`)) {
```
cd1m0, line 94: Instead of a while loop here perhaps we can use `getNewVarName(varsInScope, `RET`)`

src/instrumenter/interpose.ts
```diff
@@ -109,9 +127,13 @@ function makeStub(fun: FunctionDefinition, factory: ASTNodeFactory): FunctionDef
     let idx = 0;
 
     for (const param of stub.vParameters.vParameters) {
-        if (param.name !== "") continue;
+        if (param.name !== "") {
+            continue;
+        }
+        while (namesInFuncScope.has(`_DUMMY_ARG_${idx}`)) {
```
cd1m0, line 133: Perhaps we can use `getNewVarName` instead of a while loop here.

cd1m0 (changes requested):
Good work! I think there are still several things we can clean up though. Also I think we should emit an error when there is an `AssertionFailed` event definition already present, since we can't generated dummy names for `AssertionFailed`. (both harvey and mythril depend explicitly on naming).

Overall,  I think the code flow of scribble, and the many edge cases of naming/scoping make this PR a little tricky to implement. It may help if we add generic "scopes" to `solc-typed-ast` that solve half of the problem (what names are defined where). Perhaps if we do that, we can greatly simplify this PR.

cd1m0:
After thinking about this some more, how about we try something simpler:

1. Gather just one set with ALL the names that appear in the AST. Thus for ALL `SourceUnit`s we will accumulate:

   a) Any node with a `name` property
   b) For any ImportDirective its `unitAlias` property (if its not "")
   c) For any ImportDirective the `local` property of all its `SymbolAliases`
   
2. Given the set of all names S, make sure that anything scribble adds doesn't appear in S.

Thus in this mode we will remove `getNamesInFuncScope` and just modify and use `getAllNames`

cd1m0:
Closing in favour of https://github.com/ConsenSys/scribble/pull/13

PR #13, opened 2021-01-29
FS-25 Try 2

cd1m0:
This PR is based on work done by @norhh here https://github.com/ConsenSys/scribble/pull/11. The previous PR did a lot of good work to uncover all of the places we need to fix to handle shadowing. However, the fixes in that PR were complicated by the fact that a lot of the code they touched was not well written (by me) and over-complicated. This got in the way of fixing.

In this PR we attempt to first fix the problems in the code, and then to add code to handle the various cases of name collisions/shadowing. In detail:

1. Cleanup:
   - Convert the `InstrumentationContext` interface into its own utility class. The `InstrumentationContext` class is responsible for keeping toghether all state shared between the instrumentation of all invariants
   
   - Rename `UIDGenerator` to the better `NameGenerator` and move it inside of `InstrumentationContext` (it used to be a global var)
   
   - Add the `existingNames` field to `NameGenerator` to allow it to avoid all names that existed in the original AST
   
   - Convert several constants from global constant strings into fields of `InstrumentationContext`. Furthermore the values of these fields were obtained by calling into `NameGenerator`. This guarantees that they wont overlap with existing names, and all future instrumentation names wont conflict with them. This includes:
   
        `SCRIBBLE_VAR` -> `InstrumentationContext.structVar`
        `REENTRANCY_UTILS_CONTRACT` -> `InstrumentationContext.utilsContractName`
        `CHECK_STATE_INVS_FUN` -> `InstrumentationContext.checkStateInvsFuncName`
        `OUT_OF_CONTRACT` -> `InstrumentationContext.outOfContractFlagName`

   - Introduced the `InstrumentationContext.internalInvariantCheckers` map and corresponding function `getInternalInvariantCheckerName(ContractDefinition)` to generate names for the internal contract-invariant checker helper function without collisions.
  
   - Changed the code in `interpose(..)` to use `NameGenerator` when renaming the original functions to avoid collisions
   
   - Added the `TranspilingContext` class. It contains the state that is need to compile JUST the annotations for one function/contract. This includes a reference to `InstrumentationContext`, as well as the typing and semantic information for the group of annotations for the given function/contract.
   
    - Changed `flattenExpr` and `generateExpressions` to take a `TranspilingContext` instead of the jumble of arguments they had before
    
    - Changed the flattening and AST generation code to use `TranspilingContext` when computing field names for let-bindings, internal let-vars, internal old-vars as well as some other internals such as the scratch field (what used to be `MSTORE_SCRATCH_FIELD`) and the `CHECK_INVS_AT_END` flag.
    
    - Added tests (see `test/samples/increment_inherited_collision.sol`).


src/util/name_generator.ts
```diff
@@ -0,0 +1,39 @@
+export class NameGenerator {
+    private counters: { [base: string]: number };
```
blitz-1306, line 2: We can use the object here, ofcourse. But would be cool to have a `Map<string, number>` instead. It may have some builerplate code, but on large amount of keys it would be more efficient.

blitz-1306 (approved):
Looks good. Just a smaller suggestion.

norhh (approved):
Looks good

PR #14, opened 2021-02-05
Fixes to support re-exported imports and additional tests

cd1m0:
This pr adds support for re-exported imports and unit alias imports. E.g given the following 3 files:
```
A.sol
contract Foo {..}

B.sol
import "A.sol" as A;
import {Foo as Boo} from "A.sol";

C.sol
import {A as A1} from "B.sol";
import {Boo as Moo} from "B.sol";
import "B.sol";

contract Test{
     A1.Foo x;
     Moo y;
     Boo z;
}
```

When re-writing in flat mode, all the different aliases for `Foo` - `A1.Foo`, `Boo`, `Moo` should just be rewritten to `Foo`:

```
pragma solidity 0.7.5;

contract Foo {...}
...
contract Test {
    Foo internal x;
    Foo internal y;
    Foo internal z;
}
```

Notes:
    - this PR also fixes a bug in the grammar for imports
    - it introduces the `replaceNode` function in `scribble.ts`. That function should eventually be moved to solc-typed-ast.

### DIFF
--- a/src/ast_to_source_printer.ts
+++ b/src/ast_to_source_printer.ts
@@ -26,10 +26,11 @@ import { assert } from "./util/misc";
  * @param from - source unit from which `name` is being imported
  * @param sources - map from absolute paths to source codes
  */
-export function findImport(
+function findImport(
     name: string,
     from: SourceUnit,
-    sources: Map<string, string>
+    sources: Map<string, string>,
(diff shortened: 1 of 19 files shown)


blitz-1306 (approved):
Looks good.

PR #15, opened 2021-02-09
User functions

cd1m0:
This is a WIP PR. DO NOT MERGE.

This PR aims to add support for user-defined functions in scribble. At the moment only the grammar part is done, along with some refactoring. As it is likely gonna grow large, starting the review now. More specifically this PR:

Cleanup:

 - bumps ts-pegjs to 0.3.0 and switches it over to internally maintained fork until https://github.com/metadevpro/ts-pegjs/issues/53 gets fixed
 - renamed the type_grammar.pegjs and type_parser.ts to typeString_grammar.pegjs and typeString_parser.ts as their name was confusing
 - Re-factored expr_grammar.pegjs to include rules for the annotations themselves
 - Deprecated annotation_grammar.pegjs
 - Re-factored annotations.ts to do a single-pass parse, instead of the complicated 2-staging parsing we did before, just to emit graceful messages on missing semicolons
 
 New code:
  - Added AST nodes for declarations - SAnnotation, SProperty, SUserFunctionDefinition
  - Refactored the `Annotation` class from annotations.ts into a base `AnnotationMD` and child `PropertyMD` and `UserDefinedFunctionMd` classes. These hold a parsed definition node, as well as additional metadata needed to provide pretty errors and error line locations
  - Added test for new code in `parse.spec.ts`
  
  Note:
    - Currently there is one failing test, due to the fact that the error message for missing semicolon changed. Its a little uglier now, but on the plus side the code is much simpler. Given that in the near future these annotation comments may transition to natspec, I am willing to pay the price of a slightly uglier error message.


src/instrumenter/annotations.ts
```diff
@@ -56,75 +54,134 @@ function offsetBy(a: OffsetRange, b: number | OffsetRange): OffsetRange {
 
 let numAnnotations = 0;
 
-export class Annotation {
+/**
+ * Base class containing metadata for parsed anntotations useful for
+ * pretty-printing error messages and error line information.
+ */
+export class AnnotationMD<T extends SAnnotation> {
```
blitz-1306, line 61: I would still suggest to pick a better name than have an `MD` suffix. Even if `AnnotationMetaData` would occupy more space, it would look a bit more self-explanatory.

src/instrumenter/annotations.ts
```diff
@@ -56,75 +54,134 @@ function offsetBy(a: OffsetRange, b: number | OffsetRange): OffsetRange {
...
-        return rangeToLocRange(this.exprLoc[0], this.exprLoc[1], source);
+    annotationFileLoc(source: string): Range {
+        return rangeToLocRange(this.annotationLoc[0], this.annotationLoc[1], source);
+    }
+}
+
+/**
+ * Metadata specific to a user function definition.
+ */
+export class UserFunctionDefinitionMD extends AnnotationMD<SUserFunctionDefinition> {
```
blitz-1306, line 127: Same about this one: it would be cool to have something self-explanatory.

src/instrumenter/annotations.ts
```diff
@@ -56,75 +54,134 @@ function offsetBy(a: OffsetRange, b: number | OffsetRange): OffsetRange {
...
+        super(raw, target, original, parsedAnnot, annotationDocstringOff);
+        // Original predicate
+        this.bodyText = parsedAnnot.body.getSourceFragment(original);
+    }
+}
+
+/**
+ * Metadata specific to a property annotation (invariant, if_succeeds)
+ */
+export class PropertyMD extends AnnotationMD<SProperty> {
```
blitz-1306, line 151: Same about this one: it would be cool to have something self-explanatory.

src/instrumenter/annotations.ts
```diff
@@ -234,42 +275,27 @@ export class AnnotationExtractor {
...
             );
-        }
-
-        try {
-            const exprNode = parseExpr(exprOrig);
-
-            return new Annotation(
+        }*/
+        if (parsedAnnot instanceof SProperty) {
+            return new PropertyMD(meta.node, meta.target, annotationOrig, parsedAnnot, match.index);
```
blitz-1306, line 280: Could an "early `return`s" be used here to reduce nesting?

src/spec-lang/typeString_parser.ts
```diff
@@ -0,0 +1,5424 @@
+// Need the ts-nocheck to suppress the noUnusedLocals errors in the generated parser
```
blitz-1306, line 1: If this file is generated, then should we put it to `.gitignore` too?

blitz-1306 (commented):
Some suggestions. Generally. I think this is a good progress. Let's see how far will it go.

src/spec-lang/typeString_parser.ts
cd1m0: Fixed in 6b91a1f

src/instrumenter/annotations.ts
cd1m0: Maybe for the last else case? (`throw new Error(`NYI annotation ${parsedAnnot.pp()}`);`)? But I kinda like having all 3 on the same level here, to show that they are parts of a switch

src/instrumenter/annotations.ts
```diff
@@ -203,82 +270,45 @@ export class AnnotationExtractor {
...
-                    meta.target instanceof FunctionDefinition
-                        ? `function ${prefix}${meta.target.name}`
-                        : prefix + meta.target.name;
-            }
-
-            const errRange = rangeToLocRange(annotationLoc[0], annotationLoc[1], source);
-
-            throw new SyntaxError(
-                `Line ${errRange.start.line} of ${scope} documentation string looks like an annotation but is not terminated by a semicolon ";" and is ignored: ${annotationOrig}`,
+        if (parsedAnnot instanceof SProperty) {
```
blitz-1306, line 273: I think we can reduce nesting here via early `return`s.

src/instrumenter/instrument.ts
```diff
@@ -265,8 +284,18 @@ export function flattenExpr(expr: SNode, ctx: TranspilingContext): [SNode, SBind
...
+            if (defNode instanceof SLet) {
+                const field = ctx.getLetBinding([defNode, idx]);
+                return [getTmpVar(field, expr, expr.src), []];
+            } else if (defNode instanceof SUserFunctionDefinition) {
+                const renamedId = new SId(ctx.getUserFunArg(defNode, idx), expr.src);
+                renamedId.defSite = expr.defSite;
+                return [_registerNode(renamedId, expr), []];
+            } else {
+                throw new Error(`Unknown array def site`);
+            }
```
blitz-1306, line 298: Can we reduce nesting here by using early `return`s?

src/spec-lang/ast/declarations/annotation.ts
```diff
@@ -0,0 +1,16 @@
+import { Range, SNode } from "../node";
+export enum AnnotationType {
+    IfSucceeds = "if_succeeds",
+    Invariant = "invariant",
+    Define = "define"
```
blitz-1306, line 5: A side note (not related to the current changes): was always wondering, why do we not have `IfAborts` here, but have it in detection regex?

src/spec-lang/expr_parser_header.ts
```diff
@@ -33,11 +33,23 @@ import {
...
+    SProperty,
+    AnnotationType,
+    SUserFunctionDefinition,
+    SAnnotation
 } from "./ast";
 
 function buildBinaryExpression(head: SNode, tail: Array<[string, SNode]>, src?: Range): SNode {
     return tail.reduce((acc, [whiteSp, curOp, whiteSP, curVal]) => {
         return new SBinaryOperation(acc, curOp, curVal, src);
     }, head);
```
blitz-1306, line 46: This can be shorter if we remove curly braces and `return` construct.

src/spec-lang/tc/typecheck.ts
```diff
@@ -731,26 +796,32 @@ function tcIdVariable(expr: SId, ctx: STypingCtx, typeMap: TypeMap): SType | und
...
 
-    if (letNode.lhs.length > 1) {
-        if (!(rhsT instanceof STupleType && rhsT.elements.length === letNode.lhs.length)) {
-            throw new SExprCountMismatch(
-                `Wrong number of values for let bindings in ${letNode.pp()}. Expected ${
-                    letNode.lhs.length
-                } values, instead got ${rhsT.pp()}`,
-                letNode
-            );
+    if (node instanceof SLet) {
```
blitz-1306, line 801: Can we reduce nesting here by using early `return`s?

blitz-1306 (approved):
Looks good. A few code-related tweak suggestions to keep code-base better readable.

src/spec-lang/ast/declarations/annotation.ts
cd1m0: Just being silly. Will remove it from the detection regex.

src/instrumenter/annotations.ts
cd1m0: Fixed in 031fe1c.

src/instrumenter/instrument.ts
cd1m0: Fixed in 031fe1c.

src/spec-lang/ast/declarations/annotation.ts
cd1m0: Fixed in 031fe1c.

src/spec-lang/tc/typecheck.ts
cd1m0: Fixed in 031fe1c.

blitz-1306 (approved):

PR #17, opened 2021-02-23
Src to src mapping

cd1m0:
This PR adds a map from instrumented source locations back into original source locations. The map can be viewed in the `srcMap2SrcMap` field in `json` output mode, and in `flat` output mode as a standalone JSON object emitted with the `--scrmap-to-srcmap <file>` command line option.

## Example

For the following original code:

```solidity
pragma solidity 0.6.10;
contract Foo {
    /// if_succeeds {:msg "P0"} y == x+1;
    function foo(uint256 x) public returns (uint256 y) {
        return x + 1;
    }
}
```

The following instrumented code would be emitted:

```solidity
pragma solidity 0.6.10;

contract Foo {

    function foo(uint256 x) public returns (uint256 y) {
        y = _original_Foo_foo(x);
        if (!(y == (x + 1))) {
            emit AssertionFailed("0: P0");
            assert(false);
        }
    }

    function _original_Foo_foo(uint256 x) private returns (uint256 y) {
        return x + 1;
    }
}
/// Utility contract holding a stack counter
contract __scribble_ReentrancyUtils {
    bool __scribble_out_of_contract = true;
}
```

And the following source map would be emitted:

```json
{
  "srcMap2SrcMap": {
    "entries": [
      [
        "25:369:0",
        "24:143:0"
      ],
      ...
      [
        "373:12:0",
        "146:12:0"
      ],
      ...
    ],
    "sourceList": [
      "test/samples/increment.sol",
      "/home/dimo/work/consensys/scribble-clean/test/samples/__scribble_ReentrancyUtils.sol"
    ],
    "otherInstrumentation": [
      "482:38:0"
      "149:24:0"
    ]
  }
}
```

### Entries
The `entries` field of the object is a list of pairs of source locations, that correspond to each other. For example in the first pair - `["25:369:0", "24:143:0"]`, the first element specifies a slice of the **instrumented** code that exactly corresponds to the entire contract. The second element specifies the slice in the **original** code that corresponds to the entire contract.

Later on, in the pair `[ "373:12:0", "146:12:0"]`, the first element corresponds to the slice `return x + 1` in the instrumented file, and the second element corresponds to the slice `return x + 1` in the original code. 

In general, there is one pair of locations for (almost) every AST Node in the original code.

### Source List and file indices

The file index of the instrumented source ranges is always 0 (src2src maps are not yet supported in files mode). The file index of the original source ranges may be non-0. To be able to lookup which file it corresponds to, we add a `sourceList` field to the JSON object.

### Other Instrumentation

Not all the instrumented code maps directly to some original piece of code. Some of the code is related to the evaluation of some individual annotations. And other code is shared by many annotations and is thus 'more general'. To keep track of the additional 'shared' code, we add a `otherInstrumentation` field to the src-to-src map object. The `otherInstrumentation` field is a list of source ranges, corresponding to executable statements in the instrumented code. So in the above example the two ranges refer to the following code:

-  "482:38:0" corresponds to `bool __scribble_out_of_contract = true` - the initialization of the helper flag in ReentrancyUtils
- "149:24:0" corresponds to `y = _original_Foo_foo(x)` - the call to the original `Foo.foo` function inside of the wrapper.

### Property-specific instrumentation

This piece of code is "new" instrumentation that doesn't map directly to any original code, but is not included in `otherInstrumentation`:

```solidity
        if (!(y == (x + 1))) {
            emit AssertionFailed("0: P0");
            assert(false);
        }
```

To keep track of this code we added 2 new fields in the **property map** instead. This is what the property map looks like:

```json
{
  "propertyMap": [
    {
      "id": 0,
      "contract": "Foo",
      "filename": "test/samples/increment.sol",
      "propertySource": "71:8:0",
      "annotationSource": "47:33:0",
      "target": "function",
      "targetName": "foo",
      "debugEventSignature": "",
      "message": "P0",
      "instrumentationRanges": [
        "218:29:0",
        "218:29:0",
        "183:102:0"
      ],
      "checkRange": "187:15:0"
    }
  ]
}
```

Note the `instrumentationRanges` array - it lists all statements that were added to implement the evaluation of the property. So for example these correspond to:

- "218:29:0" - `emit AssertionFailed("0: P0")`
- "183:102:0" - 
```solidity
     if (!(y == (x + 1))) {
            emit AssertionFailed("0: P0");
            assert(false);
    }
```

As you can see this list needs to be cleaned up a little - it has duplicates and nested ranges. It can probably be collapsed and de-duplicated.

## Testing

 There is a new test harness under `src2srcmap.spec.ts`. It checks that:
 
 1. For every entry in the src2src map, the first source range corresponds to some AST node A in the instrumented code, and the second source range corresponds to some AST node B in the original code. (with some small exceptions)
 2. The nodes A and B must be of the same type (with some small exceptions)
 3. For every node in the original code we have an entry in the src2src map (with some small exceptions)


cd1m0:
This PR is now ready for review. Several major changes happened in the process of fixing since the original comment:
 
 - Support for proper source2source maps in `files` mode was added
 - The different metadata JSON objects -`propertyMap` and `src2srcMap` were unified into a single `InstrumentationMetadata` JSON object (this also involves changing the CLI options:
 - Additional `otherInstrumentation` field was added to `InstrumentationMetadata`  to keep track of new code that doesn't directly correspond to any original node or any original property
 - Property descriptions in the property map gained new fields `instrumentationRanges` and `checkRanges`. `instrumentationRanges` tracks all the new code added to evaluate a specific property. `checkRanges` tracks the exact source range, which if covered, indicates that the this property was successfully exercised.
 - Added tests that bytecode source maps are completely covered by the combination of 1) src2src maps 2) `otherInstrumentation` and 3) `instrumentationRanges` in property maps

src/instrumenter/instrumentation_context.ts
```diff
@@ -118,4 +140,24 @@ export class InstrumentationContext {
...
+    addGeneralInstrumentation(...nodes: ASTNode[]): void {
+        this.generalInstrumentationNodes.push(...nodes);
+    }
+
+    addAnnotationInstrumentation(annotation: AnnotationMetaData, ...nodes: ASTNode[]): void {
+        if (!this.evaluationStatements.has(annotation)) {
+            this.evaluationStatements.set(annotation, nodes);
+        } else {
+            (this.evaluationStatements.get(annotation) as ASTNode[]).push(...nodes);
+        }
```
blitz-1306, line 153: It would be better to remove boolean inversion (`!`) in `if` condition and swap `if`/`else` actions. Smaller amount of computations and better readability.

src/instrumenter/instrumentation_context.ts
```diff
@@ -118,4 +140,24 @@ export class InstrumentationContext {
...
+    }
+
+    addAnnotationCheck(annotation: AnnotationMetaData, pred: Expression): void {
+        if (!this.instrumetnedCheck.has(annotation)) {
+            this.instrumetnedCheck.set(annotation, [pred]);
+        } else {
+            (this.instrumetnedCheck.get(annotation) as Expression[]).push(pred);
+        }
```
blitz-1306, line 161: It would be better to remove boolean inversion (`!`) in `if` condition and swap `if`/`else` actions. Smaller amount of computations and better readability.

src/util/misc.ts
```diff
@@ -141,3 +141,25 @@ export function getScopeUnit(
 
     return node.vScope.vScope;
 }
+
+/**
+ * Given an array `arr` return a new array with all the elements of
+ * `arr` but without duplicates. Elements are compared with the `===`.
+ *
+ * @param arr
```
blitz-1306, line 149: We can remove the `@param` to keep it short.

src/bin/json_output.ts
```diff
@@ -0,0 +1,370 @@
+import {
```
blitz-1306, line 1: Can we move this file outside of `src/bin`? This directory is expected to have executable files only (with shebang prefix). Logic in this file, however is mostly utility.

test/multifile_samples/import_rewrites/instrumentationMetadata.json.expected
```diff
@@ -0,0 +1,125 @@
...
+  ],
+  "propertyMap": [],
+  "originalSourceList": [
+    "test/multifile_samples/import_rewrites/imp2.sol",
+    "test/multifile_samples/import_rewrites/imp3.sol",
+    "test/multifile_samples/import_rewrites/imp1.sol",
+    "test/multifile_samples/import_rewrites/main.sol"
+  ],
+  "instrSourceList": [
+    "--"
```
blitz-1306, line 123: Is that intentional?

test/multifile_samples/inheritance1/instrumentationMetadata.json.expected
```diff
@@ -0,0 +1,286 @@
...
+    }
+  ],
+  "originalSourceList": [
+    "test/multifile_samples/inheritance1/A.sol",
+    "test/multifile_samples/inheritance1/B.sol",
+    "test/multifile_samples/inheritance1/C.sol",
+    "test/multifile_samples/inheritance1/D.sol"
+  ],
+  "instrSourceList": [
+    "--"
```
blitz-1306, line 284: Is this intentional?

blitz-1306 (commented):
Great job! Lets polish it a bit more.

src/bin/json_output.ts
```diff
@@ -0,0 +1,370 @@
...
+
+export type InstrumentationMetaData = {
+    propertyMap: PropertyMap;
+    instrToOriginalMap: SrcToSrcMap;
+    otherInstrumentation: string[];
+    originalSourceList: string[];
+    instrSourceList: string[];
+};
+
+export type SrcTriple = [number, number, number];
```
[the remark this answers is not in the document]
cd1m0: Fixed in 6d2b8b6.

src/bin/json_output.ts
```diff
@@ -0,0 +1,370 @@
...
+}
+
+export function ppSrcTripple(src: SrcTriple): string {
+    return `${src[0]}:${src[1]}:${src[2]}`;
+}
+
+/**
+ * Returns true IFF the source range a contains the source range b.
+ * @param a
+ * @param b
```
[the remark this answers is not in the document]
cd1m0: IFF is an abbreviation meaning 'if and only if'. Fixed in 6d2b8b6.

src/instrumenter/instrumentation_context.ts
cd1m0: Fixed in 6d2b8b6.

src/util/misc.ts
cd1m0: Fixed in 6d2b8b6.

src/bin/json_output.ts
cd1m0: Moved to src/util. Fixed in 6d2b8b6.

test/multifile_samples/import_rewrites/instrumentationMetadata.json.expected
cd1m0: Yes. The "--" indicates that the output was written to stdout.

cd1m0:
It occurs to me that in JSON mode we are adding the original sources in the `source` field. Instead it might make more sense to put the instrumented sources there, and add the original sources in a new field `originalSources`

blitz-1306 (approved):
LGTM

PR #18, opened 2021-03-02
Don't emit empty invariant checkers on contracts with user functions

cd1m0:
If we have a contract with user functions but no invariants:

```
/// define one() uint = 1
contract Foo {
}
```

Scribble will currently emit a bunch of extra unnecessary state invariant checker code, due to out-of-date detection code in `scribble.ts`:

```
pragma solidity 0.8.0;

/// Utility contract holding a stack counter
contract __scribble_ReentrancyUtils {
    bool __scribble_out_of_contract = true;
}
/// define one() uint = 1
contract Foo is __scribble_ReentrancyUtils {

    /// Implementation of user function define one() uint256 = 1
    function one() internal view returns (uint256) {
        return 1;
    }

    /// Check only the current contract's state invariants
    function __scribble_Foo_check_state_invariants_internal() internal {
    }

    /// Check the state invariant for the current contract and all its bases
    function __scribble_check_state_invariants() virtual internal {
        __scribble_Foo_check_state_invariants_internal();
    }

    constructor() {
        __scribble_check_state_invariants();
        __scribble_out_of_contract = true;
    }
}
```

We should fix this to emit optimal instrumented code.

### DIFF
--- a/src/bin/scribble.ts
+++ b/src/bin/scribble.ts
@@ -222,7 +222,12 @@ function computeContractsNeedingInstr(
     contractAnnotMap: Map<ContractDefinition, PropertyMetaData[]>
 ): Set<ContractDefinition> {
     // Find the contracts needing instrumentaion by doing bfs starting from the annotated contracts
-    const wave = [...contractAnnotMap.keys()];
+    const wave = [...contractAnnotMap.entries()]
+        .filter(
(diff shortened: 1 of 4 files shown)


blitz-1306 (approved):
Looks good.

PR #20, opened 2021-03-11
Fix instrumentation metadata source index part of property values of propertySource and annotationSource

blitz-1306:
## Changes
- [x] Fix instrumentation metadata values of `propertySource` and `annotationSource` when arming was used on an independent solidity sources.

Regards.


cd1m0 (approved):
LGTM

PR #21, opened 2021-03-23
Track annotation failure actions more precisely

blitz-1306:
## Changes
- [x] Add `assertionRanges` field to instrumentation metadata to have more precise mapping of annotation failure actions.

## Notes
- Maybe it would be easier to expand https://github.com/ConsenSys/scribble/blob/b0a2395d4808796811b3dc40b290c9815fd41747/src/instrumenter/instrument.ts#L775 to `instrCtx.addAnnotationInstrumentation(annotation, ...ifStmt.getChildren(true));`? It would give close to the same effect with cleaner code.
  - [x] We [decided](https://github.com/ConsenSys/scribble/pull/21#discussion_r600061583) to leave it "as it currently is".

Regards.


test/integration/src2srcmap.spec.ts
```diff
@@ -337,8 +337,9 @@ describe("Src2src map test", () => {
                             // OR it must be part of the instrumentation of some property
                             if (
                                 forAny(instrMD.propertyMap, (prop) =>
-                                    forAny(prop.instrumentationRanges, (range) =>
-                                        contains(range, strEntry)
+                                    forAny(
```
cd1m0, line 340: `prop.assertionRanges` should be a **subset** of `prop.instrumentationRanges` so I wouldn't change this check. Instead lets add an explicit check that all `prop.assertionRanges` are contained iniside `prop.instrumentationRanges` here.

cd1m0 (approved):
LGTM. Maybe just fix the one comment about the tests?

test/integration/src2srcmap.spec.ts
cd1m0: > ## Notes
> 
>     * Maybe it would be easier to expand https://github.com/ConsenSys/scribble/blob/b0a2395d4808796811b3dc40b290c9815fd41747/src/instrumenter/instrument.ts#L775
>        to `instrCtx.addAnnotationInstrumentation(annotation, ...ifStmt.getChildren(true));`? It would give close to the same effect with cleaner code.
> 
> 
> Regards.

I would leave the code as it is, just in case some bytecode range refers to the whole if statement, instead of just its children. Then we wouldn't be covering that range.

blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/21/commits/d32f375a5afc9456b184c6a66ffb208db26c7aa2

cd1m0 (approved):
lgtm

PR #26, opened 2021-04-14
Wrap scribble computations in unchecked for Solidity >= 0.8.0

cd1m0:
After 0.8.0 Solidity arithmetic becomes 'checked' by default. In order to:

1. Have consistent behavior with Scribble pre-0.8.0
2. Ensure that spec computation does not risk throwing additional errors
3. Make it easier to write specs about functions containing unchecked computation

We have decided that scribble specs will have **unchecked** semantics. I.e. they may silently overflow without reverting.
To implement this, we wrap all scribble instrumentation-related computation in `unchecked {}` blocks.

### DIFF
--- a/test/samples/if_assigned_complex.instrumented.sol
+++ b/test/samples/if_assigned_complex.instrumented.sol
@@ -60,83 +60,101 @@ contract Base {
 
     function Base_x_inline_initializer() internal {
         vars1 memory _v;
-        _v.old_0 = x;
-        if (!(x >= _v.old_0)) {
-            emit AssertionFailed("0: ");
-            assert(false);
+        unchecked {
+            _v.old_0 = x;
+        }
+        unchecked {
+            if (!(x >= _v.old_0)) {
+                emit AssertionFailed("0: ");
+                assert(false);
+            }
         }
     }
 
     function Base_arr_idx_uint256_uint256_assign(uint256 ARG0, uint256 ARG1) internal returns (uint256 RET0) {
         arr[ARG0] = ARG1;
         RET0 = arr[ARG0];
-        if (!(arr.length > 0)) {
-            emit AssertionFailed("1: ");
-            assert(false);
-        }
-        if (!((0 <= ARG0) && (ARG0 <= arr.length))) {
-            emit AssertionFailed("2: ");
-            assert(false);
+        unchecked {
+            if (!(arr.length > 0)) {
+                emit AssertionFailed("1: ");
+                assert(false);
+            }
+            if (!((0 <= ARG0) && (ARG0 <= arr.length))) {
+                emit AssertionFailed("2: ");
+                assert(false);
+            }
         }
     }
 
     function Base_arr2_idx_uint256_idx_uint256_uint256_assign(uint256 ARG2, uint256 ARG3, uint256 ARG4) internal returns (uint256 RET1) {
         arr2[ARG2][ARG3] = ARG4;
         RET1 = arr2[ARG2][ARG3];
-        if (!(arr2.length > 0)) {
-            emit AssertionFailed("3: ");
-            assert(false);
-        }
-        if (!((0 <= ARG3) && (ARG3 <= arr2[ARG2].length))) {
-            emit AssertionFailed("5: ");
-            assert(false);
+        unchecked {
+            if (!(arr2.length > 0)) {
+                emit AssertionFailed("3: ");
+                assert(false);
+            }
+            if (!((0 <= ARG3) && (ARG3 <= arr2[ARG2].length))) {
+                emit AssertionFailed("5: ");
+                assert(false);
+            }
         }
     }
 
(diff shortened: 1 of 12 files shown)


cd1m0:
CI timed out again. It finishes fine locally. Really need to do something about this :(

test/samples/if_assigned_complex.instrumented.sol
```diff
@@ -60,83 +60,101 @@ contract Base {
 
     function Base_x_inline_initializer() internal {
         vars1 memory _v;
-        _v.old_0 = x;
-        if (!(x >= _v.old_0)) {
-            emit AssertionFailed("0: ");
-            assert(false);
+        unchecked {
+            _v.old_0 = x;
+        }
```
blitz-1306, line 65: It looks a bit strange, because I do not think if overflow/underflow may happen with such plain assignments. Is this a side effect from some changes in the code base?

test/samples/if_assigned_primitive.instrumented.sol
```diff
@@ -55,91 +55,117 @@ contract Base {
...
+            if (!(y >= _v.old_1)) {
+                emit AssertionFailed("1: ");
+                assert(false);
+            }
         }
     }
 
     function Base_x_inc_postfix() internal returns (uint256 RET3) {
         RET3 = x;
         x++;
```
blitz-1306, line 127: We can have overflow here for `++` and underflow for `--`, I guess. Should we also wrap unary operations (except `delete`) in `unchecked` block?

This is the first occurance, however there are more occurances in other snapshots. I didn't commented them all to not cause comment spamming.

blitz-1306 (commented):
Good start! I think that we need to cover a bit more edge cases:
1. We should consider user-defined function (UDF) bodies. Their call sites may be wrapped by `unchecked`, as they would be a part of predicates. However their bodies may contain literally anything that would be executed in "checked" manner as `unchecked` state is not propagated to the called function.
2. We have unary operations not wrapped by `unchecked` in few samples. I suspect that this would break the initial promise.
3. The complicated part would be if UDFs or predicates are containing a call to some local function, that is executed in checked manner. Unsure if we then can handle it properly.

Let's take a bit more time to check that these assumptions are really applicable and handle the edge cases (if there any).

test/samples/if_assigned_complex.instrumented.sol
cd1m0: This is a side effect of me just wrapping everything related to property computation in `unchecked` blocks, regardless of whether it contains arithmetic. Its just easier to write it that way.

test/samples/if_assigned_primitive.instrumented.sol
cd1m0: So this `++` corresponds to the **original** operation that we are interposing on. It is not wrapped in `unchecked` here because the original operation in `if_assigned_pritimitive.sol` was not wrapped in unchecked https://github.com/ConsenSys/scribble/blob/develop/test/samples/if_assigned_primitive.sol#L25. If you wrap line 25 in if_assigned_primitive.sol in an unchecked, then here `x++` will also be wrapped in unchecked.

cd1m0: This however brings a new problem - we may need to emit 2 versions of wrappers - one version when the wrapper is inside of an unchecked, and another when its not inside an unchecked. Good catch!

cd1m0:
> Good start! I think that we need to cover a bit more edge cases:
> 
>     1. We should consider user-defined function (UDF) bodies. Their call sites may be wrapped by `unchecked`, as they would be a part of predicates. However their bodies may contain literally anything that would be executed in "checked" manner as `unchecked` state is not propagated to the called function.
> 

Good call. I think UDF bodies should ALWAYS be inside unchecked. They are scribble code, and scribble arithmetic is by definition unchecked.

>     2. We have unary operations not wrapped by `unchecked` in few samples. I suspect that this would break the initial promise.
> 

The unary operations in the samples you are refering correspond to original operations in the solidity code, not scribble. So they should be inside an unchecked block, only IF the original expression was also in unchecked. But this also means that we may need to emit 2 versions of some wrappers. Will fix and add samples for this. Good catch!

>     3. The complicated part would be if UDFs or predicates are containing a call to some local function, that is executed in checked manner. Unsure if we then can handle it properly.
> 
I think this is ok. The "unchecked"-ness is only function-local. So if you have:

```
unchecked {
  uint x = foo(x + 1);
}

function foo(uint x) returns (uint) { return x + 1; }
```

Then the body of `foo()` is still executed as checked.

> 
> Let's take a bit more time to check that these assumptions are really applicable and handle the edge cases (if there any).

test/samples/if_assigned_primitive.instrumented.sol
cd1m0: Fixed in 0278d5c

blitz-1306 (approved):

PR #40, opened 2021-04-27
For all

norhh:
- [x] Add grammar
- [x] Add grammar tests
- [x] Add typechecks and tests
- [x] Add semantic checks and tests


src/spec-lang/ast/for_all.ts
```diff
@@ -0,0 +1,68 @@
...
+        expression: SNode,
+        start?: SNode,
+        end?: SNode,
+        start_bracket?: string,
+        end_bracket?: string,
+        array?: SNode,
+        src?: Range
+    ) {
+        super(src);
+        this.itr_type = itr_type;
```
norhh, line 28: Ignore these snake cases, will correct it.

cd1m0 (changes requested):

src/spec-lang/ast/for_all.ts
```diff
@@ -0,0 +1,70 @@
+import { SNode, Range, SId } from ".";
+import { SIntType } from "./types";
+
+/**
+ */
```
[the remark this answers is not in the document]
norhh: Done

src/spec-lang/ast/for_all.ts
```diff
@@ -0,0 +1,70 @@
...
+export class SForAll extends SNode {
+    public readonly itrType;
+    public readonly itr;
```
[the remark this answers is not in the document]
norhh: Done

src/spec-lang/ast/for_all.ts
```diff
@@ -0,0 +1,70 @@
+import { SNode, Range, SId } from ".";
+import { SIntType } from "./types";
+
+/**
+ */
+export class SForAll extends SNode {
+    public readonly itrType;
```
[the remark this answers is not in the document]
norhh: Done

test/unit/tc.spec.ts
```diff
@@ -824,7 +849,9 @@ describe("TypeChecker Annotation Tests", () => {
                 [
                     "if_assigned[bts][addr] addr == address(0x0) && bts[0] == byte(0x01);",
                     ["Unrelated", "m2"]
-                ]
+                ],
+                ["if_succeeds forall(uint i in [1...10]) arr[i] > 0;", ["Base", "plus"]],
```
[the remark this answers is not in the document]
norhh: Fixed the issue.

test/unit/tc.spec.ts
```diff
@@ -824,7 +849,9 @@ describe("TypeChecker Annotation Tests", () => {
...
+                ["if_succeeds forall(uint8 i in [a...b]) arr[i] > 0;", ["Base", "plus"]]
```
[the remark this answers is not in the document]
norhh: I cannot add the 1st one in this file, since parsing itself fails with a string.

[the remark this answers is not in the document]
cd1m0: Yeah sorry ignore the first one :)

src/spec-lang/ast/for_all.ts
```diff
@@ -0,0 +1,97 @@
...
+    public readonly iteratorVariable: SId;
+
+    /* start range */
+    public readonly start?: SNode;
+
+    /*  End range */
+    public readonly end?: SNode;
+
+    /* Start bracket type, belongs to {'[', '('} */
+    public readonly startBracket?: string;
```
cd1m0, line 25: Nit: Make type "[" | "("

src/spec-lang/ast/for_all.ts
```diff
@@ -0,0 +1,97 @@
...
+    /* End bracket type, belongs to {']', ')'} */
+    public readonly endBracket?: string;
+
+    /* condition on  which forall is applied */
+    public readonly expression: SNode;
+
+    /* Array on which iterator iterates */
+    public readonly array?: SNode;
+
+    public readonly label?: string;
```
cd1m0, line 36: Remove this one

src/spec-lang/ast/for_all.ts
```diff
@@ -0,0 +1,97 @@
...
+            return this.end;
+        }
+        return new SMemberAccess(this.array!, "length");
+    }
+
+    pp(): string {
+        if (this.start) {
+            return `(forall(${this.iteratorType} ${this.iteratorVariable.pp()} in ${
+                this.startBracket
+            } ${this.start}...${this.end} ${this.endBracket}) ${this.expression.pp()}`;
```
cd1m0, line 77: `this.start` and `this.end` need to be `this.start.pp()` and `this.end.pp()`.

src/spec-lang/ast/for_all.ts
```diff
@@ -0,0 +1,97 @@
...
+    End(): SNode {
+        if (this.end) {
+            return this.end;
+        }
+        return new SMemberAccess(this.array!, "length");
+    }
+
+    pp(): string {
+        if (this.start) {
+            return `(forall(${this.iteratorType} ${this.iteratorVariable.pp()} in ${
```
cd1m0, line 75: `this.iteratorType` needs to be `this.iteratorType.pp()`

src/spec-lang/ast/for_all.ts
```diff
@@ -0,0 +1,97 @@
...
+                this.startBracket
+            } ${this.start}...${this.end} ${this.endBracket}) ${this.expression.pp()}`;
+        } else {
+            return `(forall(${this.iteratorType} ${this.iteratorVariable.pp()} in ${
+                this.array
```
cd1m0, line 80: this.iteratorType and this.array need `.pp()`

src/spec-lang/tc/typecheck.ts
```diff
@@ -282,6 +285,12 @@ export function lookupVarDef(name: string, ctx: STypingCtx): VarDefSite | undefi
                     return [scope, i];
                 }
             }
+        } else if (scope instanceof SForAll) {
+            if (scope.iteratorVariable.name == name) {
+                return scope;
+            } else {
```
cd1m0, line 291: The `else` branch is unnecessary here

src/spec-lang/tc/typecheck.ts
```diff
@@ -1500,6 +1516,86 @@ function matchArguments(
...
+ *   - a and b should be numeric and be castable to the type of t
+ *   - type of e is boolean
+ *   - return value of this expression is boolean
+ *   - t is defined in e(t).
+ */
+export function tcForAll(expr: SForAll, ctx: STypingCtx, typeEnv: TypeEnv): TypeNode {
+    if (!(expr.iteratorType instanceof IntType)) {
+        throw new SWrongType(
+            `The expected type for ${expr.iteratorVariable.pp()} is numeric and not ${
+                expr.iteratorType
```
cd1m0, line 1531: expr.iteratorType needs `pp` here.

src/spec-lang/tc/typecheck.ts
```diff
@@ -1500,6 +1516,86 @@ function matchArguments(
...
+            }.`,
+            expr.iteratorVariable,
+            expr.iteratorType
+        );
+    }
+    assert(
+        typeof expr.start == typeof expr.end,
+        `The types of ${expr.start} and ${expr.end} are unequal, One of them is likely undefined`
+    );
+    if (expr.start && expr.end) {
```
cd1m0, line 1541: Don't need an if here - instead we should do all the checks in the if, while using the `start` and `end` getters of `ForAll`.

src/spec-lang/tc/typecheck.ts
```diff
@@ -1500,6 +1516,86 @@ function matchArguments(
...
+
+        if (!isImplicitlyCastable(expr, endT, expr.iteratorType)) {
+            throw new SWrongType(
+                `The type for ${expr.end.pp()} is not castable to ${expr.iteratorType}.`,
+                expr.end,
+                expr.iteratorType
+            );
+        }
+    }
+    if (expr.array) {
```
cd1m0, line 1577: Can remove this `if` once you convert the body of the previous if to use the `start` and `end` getter.

src/spec-lang/tc/semcheck.ts
```diff
@@ -424,3 +429,58 @@ export function scFunctionCall(
 
     throw new Error(`Internal error: NYI semcheck for ${expr.pp()}`);
 }
+
+export function sameVariable(variable1: SId, variable2: SId): boolean {
+    if (!(variable1.defSite instanceof SForAll)) {
```
cd1m0, line 434: This if is not neccessary

src/spec-lang/tc/semcheck.ts
```diff
@@ -424,3 +429,58 @@ export function scFunctionCall(
...
+ * @param semMap
+ */
+export function scForAll(expr: SForAll, ctx: SemCtx, typeEnv: TypeEnv, semMap: SemMap): SemInfo {
+    const exprSemInfo = sc(expr.expression, ctx, typeEnv, semMap);
+    const itrSemInfo = sc(expr.iteratorVariable, ctx, typeEnv, semMap);
+    const startSemInfo = sc(expr.Start(), ctx, typeEnv, semMap);
+    const endSemInfo = sc(expr.End(), ctx, typeEnv, semMap);
+    let canFail =
+        exprSemInfo.canFail || itrSemInfo.canFail || startSemInfo.canFail || endSemInfo.canFail;
+    if (expr.start && expr.end) {
```
cd1m0, line 458: Remove the if and use the getters in the body instead.

src/spec-lang/tc/semcheck.ts
```diff
@@ -424,3 +429,58 @@ export function scFunctionCall(
...
+        sc(expr.start, ctx, typeEnv, semMap);
+        sc(expr.end, ctx, typeEnv, semMap);
+    }
+    if (expr.array) {
```
cd1m0, line 462: Remove this `if`

src/spec-lang/tc/semcheck.ts
```diff
@@ -424,3 +429,58 @@ export function scFunctionCall(
...
+        const arraySemInfo = sc(expr.array, ctx, typeEnv, semMap);
+        canFail ||= arraySemInfo.canFail;
+    }
+    if (!ctx.isOld) {
+        expr.expression.walk((node) => {
+            if (
+                node instanceof SId &&
+                semMap.get(node)?.isOld &&
+                sameVariable(node, expr.iteratorVariable)
```
cd1m0, line 471: Inline `sameVariable` here and remove it.

test/unit/tc.spec.ts
```diff
@@ -646,6 +646,7 @@ describe("TypeChecker Expression Unit Tests", () => {
                 ["msg.any", ["Foo", undefined]],
                 ["tx.any", ["Foo", undefined]],
                 ["$result", ["Foo", undefined]],
+                ["$result", ["Foo", "noReturn"]],
```
cd1m0, line 649: Duplicated line?

src/spec-lang/tc/typecheck.ts
norhh: I think we can remove this part, since parser checks this.

src/spec-lang/ast/for_all.ts
```diff
@@ -0,0 +1,98 @@
...
+        this.startBracket = startBracket;
+        this.endBracket = endBracket;
+        this.expression = expression;
+        this.array = array;
+    }
+    start(): SNode {
+        if (this._start) {
+            return this._start;
+        }
+        return new SNumber(bigInt(0), 256);
```
cd1m0, line 62: Radix should be 10 not 256

src/spec-lang/ast/for_all.ts
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    getFields(): any[] {
+        return [
+            this.iteratorType,
+            this.iteratorVariable,
+            this.expression,
+            this.startBracket,
+            this.start(),
```
cd1m0, line 92: I would make those `this._start` instead of `this.start()` and `this._end` instead of `this.end()`. In general we use `getFields` to do structural equality comparison, and as such we prefer to compare the actual fields, rather than derived values of getters.

src/spec-lang/expr_grammar.pegjs
```diff
@@ -29,6 +30,31 @@ Invariant =
...
+Range = 
+    start_bracket: StartBracket __ start: Expression __ "..."  __ end: Expression __ end_bracket: EndBracket
+    {
+      return [start, end, start_bracket, end_bracket];
+    }
+    / expression: Expression {return expression;}
+    
+
+For_All = 
+  type: FORALL __ "(" __ itr_type: IntType __ iterator: Identifier __ IN __ range: Range ")" __ expr: Expression
```
cd1m0, line 48: Needs a `__` between `range: Range` and `")"`.

src/spec-lang/tc/typecheck.ts
```diff
@@ -1500,6 +1514,66 @@ function matchArguments(
...
+ * We check the following in forall (uint t in [a ... b]) e(t):
+ *   - Type of t should be numeric
+ *   - a and b should be numeric and be castable to the type of t
+ *   - type of e is boolean
+ *   - return value of this expression is boolean
+ *   - t is defined in e(t).
+ */
+export function tcForAll(expr: SForAll, ctx: STypingCtx, typeEnv: TypeEnv): TypeNode {
+    assert(
+        typeof expr.start == typeof expr.end,
```
cd1m0, line 1527: This should be `expr._start` and `expr._end`. `expr.start` and `expr.end` are just functions - they are always defined.

src/spec-lang/tc/typecheck.ts
```diff
@@ -1500,6 +1514,66 @@ function matchArguments(
...
+            expr.end(),
+            endT
+        );
+    }
+
+    if (!isImplicitlyCastable(expr, startT, expr.iteratorType)) {
+        throw new SWrongType(
+            `The type for ${expr.start().pp()} is not castable to ${expr.iteratorType}.`,
+            expr.start(),
+            expr.iteratorType
```
cd1m0, line 1553: This should be `startT`. `SWrongType` expects the actual type as the last argument.

src/spec-lang/tc/typecheck.ts
```diff
@@ -1500,6 +1514,66 @@ function matchArguments(
...
+        );
+    }
+
+    if (!isImplicitlyCastable(expr, endT, expr.iteratorType)) {
+        throw new SWrongType(
+            `The type for ${expr.end().pp()} is not castable to ${expr.iteratorType}.`,
+            expr.end(),
+            expr.iteratorType
```
cd1m0, line 1561: This should be `endT`. `SWrongType` expects the actual type as the last argument.

test/unit/parser.spec.ts
```diff
@@ -572,7 +574,40 @@ describe("Expression Parser Unit Tests", () => {
...
+                    new SIndexAccess(new SId("a"), new SId("x")),
+                    ">",
+                    new SNumber(bigInt(10), 10)
+                ),
+                new SNumber(bigInt(1), 10),
+                new SNumber(bigInt(10), 10),
+                "[",
+                ")"
+            )
+        ],
```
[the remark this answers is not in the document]
cd1m0: We are still missing a test case where start and end are complex expressions

test/unit/tc.spec.ts
```diff
@@ -850,7 +853,30 @@ describe("TypeChecker Annotation Tests", () => {
                     new BoolType(),
                     true
                 ],
-                ["if_updated old(z)>0;", ["Unrelated", "z"], new BoolType(), true]
```
cd1m0, line 853: The existing test for `if_udpated old(z)>0;` was deleted here

test/unit/parser.spec.ts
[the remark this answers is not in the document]
cd1m0: Nevermind - there is a test in tc.spec.ts that also covers this.

src/spec-lang/tc/semcheck.ts
```diff
@@ -424,3 +429,45 @@ export function scFunctionCall(
...
+ */
+export function scForAll(expr: SForAll, ctx: SemCtx, typeEnv: TypeEnv, semMap: SemMap): SemInfo {
+    const exprSemInfo = sc(expr.expression, ctx, typeEnv, semMap);
+    const itrSemInfo = sc(expr.iteratorVariable, ctx, typeEnv, semMap);
+    const startSemInfo = sc(expr.start(), ctx, typeEnv, semMap);
+    const endSemInfo = sc(expr.end(), ctx, typeEnv, semMap);
+    const canFail =
+        exprSemInfo.canFail || itrSemInfo.canFail || startSemInfo.canFail || endSemInfo.canFail;
+
+    sc(expr.start(), ctx, typeEnv, semMap);
```
cd1m0, line 449: We are calling `sc` on `expr.start()` and `expr.end()` twice here. This is unnecessary.

src/spec-lang/tc/semcheck.ts
```diff
@@ -424,3 +429,45 @@ export function scFunctionCall(
...
+    sc(expr.end(), ctx, typeEnv, semMap);
+
+    if (!ctx.isOld) {
+        expr.expression.walk((node) => {
+            if (
+                node instanceof SId &&
+                semMap.get(node)?.isOld &&
+                node.defSite instanceof SForAll &&
+                node.defSite.iteratorVariable.id == expr.iteratorVariable.id
```
cd1m0, line 458: You can replace

```
               node.defSite instanceof SForAll &&
               node.defSite.iteratorVariable.id == expr.iteratorVariable.id
```

with just `node.defSite === expr`.

src/spec-lang/tc/semcheck.ts
```diff
@@ -427,3 +432,31 @@ export function scFunctionCall(
...
+
+/**
+ * For Any expression of form old(e1) in `forall(type t in range) e(t)`
+ * throw an error if the expression depends on t.
+ * @param expr
+ * @param ctx
+ * @param typeEnv
+ * @param semMap
+ */
+export function scForAll(expr: SForAll, ctx: SemCtx, typeEnv: TypeEnv, semMap: SemMap): SemInfo {
```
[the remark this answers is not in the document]
cd1m0: So this case worries me a little. It might just work, but I think we should add the following integration tests:

``` forall (uint i in old(a1)) ...```
``` forall (uint i in old(a2[0])) ... ```  (where a2 is a 2 dimensional array)
``` forall (uint i in old(a)) forall (uint j in old(a[i])) ....``` (where a is a 2-dimensional array)

If these compile and run we can leave this without a check. Otherwise we should go back and add checks that the ranges/array are not inside an `old()`.

This is not a blocker to merge this PR.

**Review by cd1m0 — CHANGES_REQUESTED** [review]
Good work! Some nits left.

File: src/spec-lang/tc/typecheck.ts (hunk 24 above)
Comment (norhh) [reply src/spec-lang/tc/typecheck.ts:1527]: We can remove this assert now right?

**Comment by norhh** [comment]
Currently I cannot run forall in integration tests as there is no transpiling implementation for `ForAll` node. Should I add those to semcheck.ts?

File: src/spec-lang/tc/typecheck.ts (hunk 24 above)
Comment (cd1m0) [reply src/spec-lang/tc/typecheck.ts]: Yes. Though it would be good to add a special-case error message here for the case `forall (type i in a)` for when `a` is not an array, and for when `type` is not a uint. Otherwise the error messages may be confusing, as they will refer to `a.length` which not present in the code. This is not a blocker for this PR, should be fixed in the next

**Review by cd1m0 — APPROVED** [review]
Small nits to fix in the next PR. Ready to merge

File: src/spec-lang/tc/typecheck.ts (hunk 24 above)
Comment (cd1m0) [reply src/spec-lang/tc/typecheck.ts]: Fixed.

**Comment by cd1m0** [comment]
1. Added the instrumentation part and test

2. After conversations with ppl simplified the syntax for ranges from "[a...b]" to "a...b".

3. Added a bunch of tests

File: test/samples/forall_simple.sol [hunk 32]
```diff
@@ -0,0 +1,51 @@
...
+
+	/// #if_succeeds forall (uint i in 1...5) i >= 1 && i < 5;
+	function test5() public {
+	}
+
+	/// #if_succeeds forall (uint i in 1...5) i > 1;
+	function test6() public {
+	}
+
+	/// #if_succeeds forall (uint i in 1...5) i < 4;
```
wuestholz, line 40: @cd1m0 Is this supposed to hold?

test/samples/forall_simple.sol
```diff
@@ -0,0 +1,51 @@
...
+	/// #if_succeeds forall (uint i in a) forall (uint j in a[i]) a[i][j] > old(t2);
+	function test4(uint[][] memory a, uint x) public {
+		t2 = x;
+	}
+
+	/// #if_succeeds forall (uint i in 1...5) i >= 1 && i < 5;
+	function test5() public {
+	}
+
+	/// #if_succeeds forall (uint i in 1...5) i > 1;
```
wuestholz, line 36: @cd1m0 Is this supposed to hold?

test/samples/forall_simple.instrumented.sol
```diff
@@ -0,0 +1,355 @@
...
+            _v.old_2 = t2;
+        }
+        _original_ForallSimple_test4(a, x);
+        unchecked {
+            _v.forall_4 = true;
+            for (_v.i4 = 0; _v.i4 < a.length; _v.i4++) {
+                _v.forall_5 = true;
+                for (_v.j0 = 0; _v.j0 < a[_v.i4].length; _v.j0++) {
+                    _v.forall_5 = a[_v.i4][_v.j0] > _v.old_2;
+                    if (!_v.forall_5) break;
```
wuestholz, line 199: @cd1m0 I assume in Solidity this breaks only out of the inner loop. Right?

src/spec-lang/expr_grammar.pegjs
```diff
@@ -130,6 +151,7 @@ IF_SUCCEEDS = "if_succeeds"
 IF_UPDATED = "if_updated"
 IF_ASSIGNED = "if_assigned"
 DEFINE = "define"
+FORALL = "forall"
```
blitz-1306, line 154: Should we also add it to `Keyword` rule?

src/spec-lang/expr_grammar.pegjs
```diff
@@ -37,6 +38,26 @@ Invariant =
...
+For_All =
+  type: FORALL __ "(" __ itr_type: IntType __ iterator: Identifier __ IN __ range: Range __ ")" __ expr: Expression
+  {
+    if(Array.isArray(range)) {
+      const [start, end] = range;
+      return new SForAll(itr_type, iterator, expr, start, end, undefined, location());
+    }
+    else {
+      return new SForAll(itr_type, iterator, expr, undefined, undefined, range, location());
+    }
```
blitz-1306, line 58: We can reduce nesting here. On the other hand, I can do a full cleanup pass for the grammar in other PR (if you do not mind).

src/spec-lang/tc/semcheck.ts
```diff
@@ -424,3 +429,37 @@ export function scFunctionCall(
...
     throw new Error(`Internal error: NYI semcheck for ${expr.pp()}`);
 }
+
+/**
+ * For Any expression of form old(e1) in `forall(type t in range) e(t)`
+ * throw an error if the expression depends on t.
+ * @param expr
+ * @param ctx
+ * @param typeEnv
+ * @param semMap
```
blitz-1306, line 439: Params can be removed as there is no info specified for their description.

test/samples/forall_simple.instrumented.sol
blitz-1306: @wuestholz It will cause a cascade of `break`s:

```solidity
_v.forall_5 = a[_v.i4][_v.j0] > _v.old_2;
if (!_v.forall_5) break;

// ...

_v.forall_4 = _v.forall_5;
if (!_v.forall_4) break;

// ...

if (!(_v.forall_4)) {
    emit AssertionFailed("4: ");
    assert(false);
}
```

I guess we better state such behavior in the docs (I mean nested `forall`s).

test/samples/forall_simple.sol
blitz-1306: Good notice. I'm wondering if behavior here should be same for array traversal and fixed range. For fixed range it is would be obvious to include `1` and `5`. For array traversal, last element should be less than `array.length`.

If this supposed to hold, I'm debating that fixed ranges are suffering a bit.

blitz-1306 (commented):
Code base looks clean. Few questions and suggestions.

test/samples/forall_simple.instrumented.sol
wuestholz: 👍 Right, that's also what I assumed. I was worried about the semantics of nested breaks since, as far as I remember, some languages automatically break out of the outer loops as well.

test/samples/forall_simple.sol
wuestholz: @blitz-1306 Are you saying that `1...5` "feels" like 5 would be included? If so, I kind of agree and we probably need to stress this in the docs.

test/samples/forall_simple.sol
cd1m0: No - this is supposed to revert. There is a test in https://github.com/ConsenSys/scribble/pull/40/files#diff-615b432e8bcc686549a508d00ebfdf8bf8c2f846a9cbc0bb1a2a13c0c9e963d0 that checks that when calling `test7` after instrumentation it reverts.

test/samples/forall_simple.sol
cd1m0: No - this is supposed to revert. There is a test in https://github.com/ConsenSys/scribble/pull/40/files#diff-615b432e8bcc686549a508d00ebfdf8bf8c2f846a9cbc0bb1a2a13c0c9e963d0 that checks that when calling `test6` after instrumentation it reverts.

cd1m0: This test is just here to make sure the instrumentation  really is checking the ranges strictly.

test/samples/forall_simple.sol
cd1m0: This test is just here to make sure the instrumentation  really is checking the ranges strictly.

test/samples/forall_simple.instrumented.sol
cd1m0: Solidity `break` breaks only out of the innermost loop.

src/spec-lang/tc/semcheck.ts
cd1m0: Fixed in 2c86523

src/spec-lang/expr_grammar.pegjs
cd1m0: Fixed in 2c86523

cd1m0 (approved):
lgtm

cd1m0:
@blitz-1306 @wuestholz  any objections to merging this?

blitz-1306 (approved):
No objections from me. Good to go.

Issue #39, opened 2021-04-23
Bug if using if_updated and if_succeeds


newcodepusher:
scribble 0.4.2

``` solidity
pragma solidity >=0.6.0 <0.8.0;

contract Token {
    function balanceOf(address x) external pure returns (uint) {
        return 123;
    }
}

contract Foo {
    /// if_updated tokenBalance == 123;
    uint public tokenBalance = 1;
    
    /// if_succeeds Token(x).balanceOf(address(this)) == 123;
    function updateBalance(address x) public returns (uint) {
        tokenBalance = Token(x).balanceOf(address(this));
        return tokenBalance;
    }
}

```

```
/.nvm/versions/node/v12.22.1/lib/node_modules/eth-scribble/dist/bin/scribble.js:574
            throw e;
            ^

Error: Mismatch in length between [] of len 0 and [Identifier#81] of len 1
    at assert (/.nvm/versions/node/v12.22.1/lib/node_modules/eth-scribble/dist/util/misc.js:16:11)
    at Object.zip (/.nvm/versions/node/v12.22.1/lib/node_modules/eth-scribble/dist/util/misc.js:194:5)
    at helper (/.nvm/versions/node/v12.22.1/lib/node_modules/eth-scribble/dist/instrumenter/state_vars.js:99:24)
    at getAssignments (/.nvm/versions/node/v12.22.1/lib/node_modules/eth-scribble/dist/instrumenter/state_vars.js:204:20)
    at getAssignments.next (<anonymous>)
    at Object.findAliasedStateVars (/.nvm/versions/node/v12.22.1/lib/node_modules/eth-scribble/dist/instrumenter/state_vars.js:262:21)
    at instrumentFiles (/.nvm/versions/node/v12.22.1/lib/node_modules/eth-scribble/dist/bin/scribble.js:165:38)
    at Object.<anonymous> (/.nvm/versions/node/v12.22.1/lib/node_modules/eth-scribble/dist/bin/scribble.js:568:40)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)

```

cd1m0:
Thank you for reporting this! Looking at it now

cd1m0:
Ok I think I found it. The issue is on this line:

https://github.com/ConsenSys/scribble/blob/6e737c9ff7db2e0466825411b86a810a6079f988/src/instrumenter/transpile.ts#L387

[...]

PR #41, opened 2021-05-04
Remove old type string parser and update Scribble to use solc-typed-ast

cd1m0:
We migrated the logic to infer the type of an `Expression`/`VariableDeclaration` ASTNode from Scribble into solc-typed-ast 5.0.1. This PR removes the redundant code from scribble, and updates it to use soclt-typed-ast 5.0.1


src/instrumenter/state_var_instrumenter.ts
```diff
@@ -205,45 +202,45 @@ const assignSuffixMap: { [key: string]: string } = {
 /**
  * Get a string descriptor for a type to be used in naming conventions
  */
-function getTypeDescriptor(typ: SType): string {
-    if (typ instanceof SAddressType) {
+function getTypeDescriptor(typ: TypeNode): string {
```
blitz-1306, line 205: I'm wondering if we can reuse `TypeNode`s `pp()` method. There a few differences here, though. And I guess it serves a bit different purpose.

src/spec-lang/tc/builtins.ts
```diff
@@ -1,143 +1,178 @@
...
+    pp(): string {
+        return `builtin_struct ${this.name}`;
+    }
+
+    getFields(): any[] {
+        return [this.name];
+    }
+}
+
+export const BuiltinSymbols = new Map<string, TypeNode>([
```
blitz-1306, line 34: Missing `bytes.concat` here I guess (introduced in [Solidity 0.8.4](https://github.com/ethereum/solidity/releases/tag/v0.8.4))

src/spec-lang/tc/builtins.ts
```diff
@@ -1,143 +1,178 @@
...
+            [new AddressType(false)],
             FunctionVisibility.Default,
             FunctionStateMutability.Pure
         )
     ]
 ]);
 
-export const BuiltinAddressMembers = new Map<string, SType>([
-    ["balance", new SIntType(256, false)],
+export const BuiltinAddressMembers = new Map<string, TypeNode>([
```
blitz-1306, line 168: Missing `address.code` here I guess.

src/spec-lang/tc/typecheck.ts
```diff
@@ -938,29 +775,29 @@ function locateElementType(type: TypeName, path: DatastructurePath): SType {
...
     }
 
-    if (idxCompT.to instanceof SArrayType) {
-        return new SIntType(256, false);
-    } else if (idxCompT.to instanceof SMappingType) {
+    if (idxCompT.to instanceof ArrayType) {
+        return new IntType(256, false);
+    } else if (idxCompT.to instanceof MappingType) {
         return idxCompT.to.keyType;
     } else {
```
blitz-1306, line 802: I guess we can reduce nesting here.

src/spec-lang/tc/typecheck.ts
```diff
@@ -1788,14 +1628,14 @@ export function tcFunctionCall(expr: SFunctionCall, ctx: STypingCtx, typeEnv: Ty
 
         if (def instanceof FunctionDefinition) {
             // param.vType is defined, as you can't put a `var x,` in a function definition.
-            const retTs = def.vReturnParameters.vParameters.map((param) => astVarToSType(param));
+            const retTs = def.vReturnParameters.vParameters.map((param) => astVarToTypeNode(param));
```
blitz-1306, line 1631: Can just pass `astVarToTypeNode` instead of creating another closure `(param) => astVarToTypeNode(param)` ?

blitz-1306 (commented):
Good work. A few suggestions.

src/instrumenter/state_var_instrumenter.ts
cd1m0: They output differently formatted strings. TypeNode's `pp()` method outputs human-readable strings with spaces. This method outputs strings with no spaces that are used for computing wrapper names

cd1m0: I don't think we can re-use that here

src/spec-lang/tc/builtins.ts
cd1m0: `bytes.concat` is difficult to add because it supports a variable number of arguments. Supporting variable-number-of-arguments builtins requires re-architecting some stuff here. Since this is larger, lets do it as a separate PR

src/spec-lang/tc/builtins.ts
cd1m0: I added it but there is an interesting problem here - scribble will now accept `address().code` on ALL solidity versions (including <0.8.0). And we will produce code that doesn't compile. So it seems that we need to modify this datastructure to make it aware of what versions ranges are supported for each.

src/spec-lang/tc/typecheck.ts
cd1m0: That unfortunately doesn't work due to the extra optional parameters for `astVarTyoTypeNode`. They conflict with the optional parameters for `map` closures.

src/spec-lang/tc/typecheck.ts
cd1m0: Fixed in f4281ca

src/spec-lang/tc/builtins.ts
cd1m0: Fixed in f4281ca.

blitz-1306 (approved):

src/instrumenter/transpile.ts
```diff
@@ -1,90 +1,87 @@
...
 
-    if (type instanceof SPointer) {
+    if (type instanceof PointerType) {
         return generateTypeAst(type.to, factory);
     }
 
-    if (type instanceof SUserDefinedType) {
+    if (type instanceof UserDefinedType) {
         // @todo remove this hack when we fix the types obtained from the typeString parser in getExprSType()
         const id = type.definition !== undefined ? type.definition.id : -1;
```
cd1m0, line 80: Remove this hack and comment.

src/instrumenter/transpile.ts
```diff
@@ -308,9 +303,9 @@ export function generateExprAST(
 
         let referencedDeclaration = -1;
 
-        if (type instanceof SFunctionSetType) {
+        if (type instanceof FunctionSetType) {
             referencedDeclaration = type.definitions[0].id;
```
cd1m0, line 307: Maybe it would be better if this was `single(type.definitions).id`. Who is responsible for guaranteeing overloaded functions in the set are filtered here? Or is this done only at the parent level (FunctionCall)?

src/spec-lang/expr_parser_header.ts
```diff
@@ -41,16 +30,64 @@ import {
...
+import { assert } from "../util/misc";
 
 function buildBinaryExpression(head: SNode, tail: Array<[string, SNode]>, src?: Range): SNode {
     return tail.reduce((acc, [whiteSp, curOp, whiteSP, curVal]) =>
         new SBinaryOperation(acc, curOp, curVal, src), head);
 }
 
-export function parseAnnotation(str: string): SAnnotation {
+export function parseAnnotation(str: string, ctx: ASTNode, version: string): SAnnotation {
     return parse(str, { startRule: "Annotation"});
```
cd1m0, line 60: Need to provide ctx and version here. How is this even passing tests?

src/spec-lang/tc/builtins.ts
```diff
@@ -1,145 +1,181 @@
...
+    IntType,
+    PointerType,
+    TypeNode,
+    Range,
+    BuiltinType
+} from "solc-typed-ast";
 
-export const BuiltinSymbols = new Map<string, SType>([
-    ["abi", new SBuiltinStructType("abi", new Map())],
+export class BuiltinStructType extends BuiltinType {
```
cd1m0, line 17: Move class to own file

src/spec-lang/tc/typecheck.ts
```diff
@@ -130,6 +123,35 @@ function getScopeOfType<T extends ContractDefinition | FunctionDefinition>(
     return undefined;
 }
 
+/**
+ * Internal class used during type-checking to carry information for the possible overloaded targets
+ * of a callee, back to the callsite which can resolve based on the actual arguments.
+ */
+export class FunctionSetType extends TypeNode {
```
cd1m0, line 130: Move to own file

src/spec-lang/tc/typecheck.ts
```diff
@@ -1026,20 +865,17 @@ function tcIdVariable(expr: SId, ctx: STypingCtx, typeEnv: TypeEnv): SType | und
...
-        return new SPointer(
-            new SUserDefinedType(getUserDefinedTypeFQName(contract), contract),
-            DataLocation.Storage
-        );
+        return mkUserDefinedType(contract);
     }
 
     // First try to TC the id as a builtin type
-    let retT: SType | undefined = tcIdBuiltinType(expr);
+    let retT: TypeNode | undefined = tcIdBuiltinType(expr);
```
cd1m0, line 878: Can the rest of this function, and the tcIdVariable() functions be signifficantly refactored using the new `resolveAny` function?

src/spec-lang/tc/typecheck.ts
```diff
@@ -1118,80 +955,81 @@ export function tcResult(expr: SResult, ctx: STypingCtx, typeEnv: TypeEnv): STyp
...
+    if (type instanceof IntLiteralType && to instanceof IntType) {
         return true;
     }
 
-    if (type instanceof SStringLiteralType && to instanceof SPointer && to.to instanceof SBytes) {
+    if (
+        type instanceof StringLiteralType &&
+        to instanceof PointerType &&
+        to.to instanceof BytesType
+    ) {
```
cd1m0, line 1011: This if and the one below can be unified. This function needs comments too

src/spec-lang/tc/typecheck.ts
```diff
@@ -1569,50 +1408,62 @@ export function tcMemberAccess(expr: SMemberAccess, ctx: STypingCtx, typeEnv: Ty
...
 
     if (
-        baseT instanceof SUserDefinedTypeNameType &&
-        baseT.definition instanceof ContractDefinition
+        baseT instanceof TypeNameType &&
+        baseT.type instanceof UserDefinedType &&
+        baseT.type.definition instanceof ContractDefinition
     ) {
         // First check if this is a type name
-        const type = resolveTypeDef([ctx[0], baseT.definition], expr.member);
```
cd1m0, line 1576: This if could also be simplified using resolveAny. This requires passing the version to `tc()` somehow.

src/spec-lang/tc/typecheck.ts
```diff
@@ -1711,40 +1563,30 @@ export function tcFunctionCall(expr: SFunctionCall, ctx: STypingCtx, typeEnv: Ty
...
-        let loc: DataLocation;
 
         if (calleeT.definition instanceof StructDefinition) {
-            // Struct constructor case
-            loc = DataLocation.Memory;
-        } else {
-            // Type-casting case
+            // Struct constructor case - always lives in memory
+            return new PointerType(mkUserDefinedType(calleeT.definition), DataLocation.Memory);
+        } else if (calleeT instanceof ContractDefinition || calleeT instanceof EnumDefinition) {
```
cd1m0, line 1573: Unnest here.

src/util/misc.ts
```diff
@@ -285,3 +285,7 @@ export function updateMap<T1, T2>(
         updatee.set(key, val);
     }
 }
+
+export function last<T>(arr: T[]): T {
```
cd1m0, line 289: This is not used anywhere. There are several places it can be used. Also assert `arr.length > 0` inside.

test/unit/interposing.spec.ts
```diff
@@ -319,8 +319,8 @@ contract Foo {
     ] of goodSamples) {
         it(`Interpose on ${contractName}.${funName} in #${fileName}`, () => {
             const [sources, reader, files, compilerVersion] = toAst(fileName, content);
-            const typeCtx = getTypeCtx([contractName, funName], sources);
-            const fun: FunctionDefinition = getTarget(typeCtx) as FunctionDefinition;
+            const [, target] = getTypeCtxAndTarget([contractName, funName], sources);
```
cd1m0, line 322: Use `getTarget` instead of `getTypeCtxAndTarget`

test/unit/interposing.spec.ts
```diff
@@ -441,8 +441,8 @@ contract Foo is __scribble_ReentrancyUtils {
         it(`Instrument ${contractName} in #${fileName}`, () => {
             const [sources, reader, files, compilerVersion] = toAst(fileName, content);
 
-            const typeCtx = getTypeCtx([contractName, undefined], sources);
-            const contract: ContractDefinition = getTarget(typeCtx) as ContractDefinition;
+            const [, target] = getTypeCtxAndTarget([contractName, undefined], sources);
```
cd1m0, line 444: Use `getTarget` instead of `getTypeCtxAndTarget`

src/instrumenter/state_var_instrumenter.ts
```diff
@@ -276,12 +274,22 @@ function getWrapperName(
             }
             const [expectedTyp, expr] = el;
 
-            const exprT = getExprSType(expr, astTypeNameToSType(expectedTyp));
+            const exprT = getMaterialExprType(
```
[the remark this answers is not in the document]
cd1m0: A: Because otherwise we have to assume some default values for storage location for complex key types (bytes/strings in mappigns) and additional arguments (values we are assigning). If we assume default types we may introduce extra deep-copies in the code. The tradeoff we are picking here is to emit multiple specialized versions of the same wrapper depending on arguements. Also this way we can avoid dealing with implicit casts from array literals to more general storage arrays.

src/instrumenter/state_var_instrumenter.ts
```diff
@@ -276,12 +274,22 @@ function getWrapperName(
...
+                version,
+                updateNode,
+                typeNameToTypeNode(expectedTyp)
+            );
             return `idx_${getTypeDescriptor(exprT)}`;
         })
         .join("_");
+
     const additionalArgsString = additionalArgs
-        .map(([expr, typ]) => getTypeDescriptor(getExprSType(expr, astTypeNameToSType(typ))))
```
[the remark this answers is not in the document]
cd1m0: A: Because otherwise we have to assume some default values for storage location for complex key types (bytes/strings in mappigns) and additional arguments (values we are assigning). If we assume default types we may introduce extra deep-copies in the code. The tradeoff we are picking here is to emit multiple specialized versions of the same wrapper depending on arguements. Also this way we can avoid dealing with implicit casts from array literals to more general storage arrays.

src/instrumenter/state_var_instrumenter.ts
```diff
@@ -413,13 +421,23 @@ function makeWrapper(
...
+            ctx.compilerVersion,
+            updateNode,
+            typeNameToTypeNode(keyT)
+        );
         formalParamTs.push(exprT);
     }
 
     for (const [actual, formalT] of additionalArgs) {
         replMap.set(actual.id, formalParamTs.length);
-        const exprT = getExprSType(actual, astTypeNameToSType(formalT));
```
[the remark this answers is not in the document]
cd1m0: A: Because otherwise we have to assume some default values for storage location for complex key types (bytes/strings in mappigns) and additional arguments (values we are assigning). If we assume default types we may introduce extra deep-copies in the code. The tradeoff we are picking here is to emit multiple specialized versions of the same wrapper depending on arguements. Also this way we can avoid dealing with implicit casts from array literals to more general storage arrays.

src/instrumenter/transpile.ts
cd1m0: Ensuring that there is a single definition is done in `tcFunctionCall` here: https://github.com/ConsenSys/scribble/blob/f4281ca2b31329f988487e14c4b77e1ab10357d7/src/spec-lang/tc/typecheck.ts#L1629

src/spec-lang/tc/typecheck.ts
cd1m0: Not quite. We can't replace `tcIdVariable` with `resolveAny` as `tcIdVariable` incorporates non-solidity definitions (let-bindings, user-defined functions and bindings in if_updated paths). We could replace the calls to `resolveByName` and `resolveTypeDef`, but we would need to propagate version and context into `tc()` which is a bunch of extra work. We should probably do this in the future though.

src/instrumenter/state_var_instrumenter.ts
```diff
@@ -413,13 +421,23 @@ function makeWrapper(
 
         const [keyT, keyExp] = pathEl;
         replMap.set(keyExp.id, formalParamTs.length);
-        const exprT = getExprSType(keyExp, astTypeNameToSType(keyT));
```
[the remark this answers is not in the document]
cd1m0: A: Because otherwise we have to assume some default values for storage location for complex key types (bytes/strings in mappigns) and additional arguments (values we are assigning). If we assume default types we may introduce extra deep-copies in the code. The tradeoff we are picking here is to emit multiple specialized versions of the same wrapper depending on arguements. Also this way we can avoid dealing with implicit casts from array literals to more general storage arrays.

src/spec-lang/tc/typecheck.ts
cd1m0: Again this should be done as follow-up cleanup, due to the effort of propagating version and context throughout the `tc()` code.

src/spec-lang/tc/typecheck.ts
cd1m0, line 1573: This is buggy - the if condition should be `calleeT.definition instanceof ContractDefinition || calleeT.definition instanceof EnumDefinition`. Again its weird that tests didn't catch this. Perhaps we need more casting tests?

src/instrumenter/transpile.ts
cd1m0: Fixed in ef446bd.

src/instrumenter/transpile.ts
cd1m0: Fixed in ef446bd.

src/spec-lang/expr_parser_header.ts
cd1m0: Fixed in ef446bd.

src/spec-lang/tc/builtins.ts
cd1m0: Fixed in ef446bd.

src/spec-lang/tc/typecheck.ts
cd1m0: Fixed in ef446bd.

src/spec-lang/tc/typecheck.ts
cd1m0: Fixed in ef446bd.

src/spec-lang/tc/typecheck.ts
cd1m0: Fixed in ef446bd.

src/util/misc.ts
cd1m0: Fixed in ef446bd.

test/unit/interposing.spec.ts
cd1m0: Fixed in ef446bd.

test/unit/interposing.spec.ts
cd1m0: Fixed in ef446bd.

src/spec-lang/tc/typecheck.ts
cd1m0: Fixed in ef446bd.

cd1m0:
Note: This PR now includes a fix for https://github.com/ConsenSys/scribble/issues/39 as well.

src/util/misc.ts
```diff
@@ -285,3 +285,8 @@ export function updateMap<T1, T2>(
...
+    assert(arr.length > 0, `Internal Error: Unexpected empty array`);
```
blitz-1306, line 290: I guess we better warn developer(s) that `arr` s expected to be non-empty (`length > 0`). I generally like the idea of shorthand, however I think that it sometimes may be  better to delegate error handling a logic at particular callsite (because of possible edge cases) without using `try`/`catch` or workarounds.

Now I would suggest something that I would not do quite often: add a boolean flag `throwOnEmpty = true` and consider overloading.
```typescript
export function last<T>(arr: T[]): T;
export function last<T>(arr: T[], throwOnEmpty: true): T;
export function last<T>(arr: T[], throwOnEmpty: false): T | undefined;
export function last<T>(arr: T[], throwOnEmpty = true): T | undefined {
    assert(!throwOnEmpty || arr.length > 0, `Internal Error: Unexpected empty array`);

    return arr[arr.length - 1];
}
```
We can also think if the flag is `allowEmpty` instead if you prefer, but idea is to derive the checks to the type system and not cause confusion / accidental bugs. Other thing would be to have two different implementations (one with `assert()` call and one without it).

blitz-1306 (approved):
Thanks for the better comments about type casting and other fixed. A small suggestion.

blitz-1306 (approved):

src/util/misc.ts
blitz-1306: Fixed in https://github.com/ConsenSys/scribble/pull/41/commits/7dfdbafe0d9b836e1e7ab083c07dc4eb3bdfa24c

PR #43, opened 2021-05-05
Bugfix toJSNumber

norhh:
apparently toJSNumber is not a function.


cd1m0:
I'm confused by this PR. In the grammar where we use toJSNumber the target is not the native BigInt type, its the BigInteger object from https://www.npmjs.com/package/big-integer, and as far as I can tell that object does have a toJSNumber function: https://www.npmjs.com/package/big-integer#tojsnumber. This was older code from before we had moved on to a nodejs version that supported native BigInts. Also if `toJSNumber` wasn't a function, a lot more tests should have failed?

cd1m0 (approved):
Good job catching this!

PR #44, opened 2021-05-12
Support for `#` prefix

blitz-1306:
## Changes
- [x] Add support for marking annotations with `#` character. It helps to properly track the annotation beginning for annotation types of `invariant` or `define` (these can be used as a regular words). This also would simplify life for the tool developers as annotations would have begin (`#`) and end (`;`) markers.
- [x] Added deprecation notice for annotations without `#` prefix.
- [x] Added relevant test file and tweaked existing tests to not fail due to a deprecation notice.

Regards.


src/instrumenter/annotations.ts
```diff
@@ -395,7 +395,7 @@ class AnnotationExtractor {
 
         const result: AnnotationMetaData[] = [];
 
-        const rx = /\s*(\*|\/\/\/)\s*(if_succeeds|if_updated|if_assigned|invariant|define\s*[a-zA-Z0-9_]*\s*\([^)]*\))/g;
+        const rx = /\s*(\*|\/\/\/)(\s*#)?\s*(if_succeeds|if_updated|if_assigned|invariant|define\s*[a-zA-Z0-9_]*\s*\([^)]*\))/g;
```
cd1m0, line 398: I don't think we should allow space between the `#` and the keyword. (So no `#  if_succeeds`)

src/spec-lang/expr_grammar.pegjs
```diff
@@ -1,6 +1,6 @@
 // Top-level rules
 Annotation
-    = StartingWhiteSpace annotation: (Invariant
+    = StartingWhiteSpace ('#' __)? annotation: (Invariant
```
cd1m0, line 3: Same here - remove the '__' after '#'

cd1m0 (changes requested):
Good work! Just two things:

1. Fix the space thing after #

2. Lets tweak this PR so that when we encounter a correctly parsed annotation without the `#` we emit a warning to the user that "Annotations without `#` will be eventually deprecated. Please add a `#` to your annotations"

src/spec-lang/expr_grammar.pegjs
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/44/commits/02b98f0d598cab8ac5fc53764fc058c3f88404a8

src/instrumenter/annotations.ts
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/44/commits/02b98f0d598cab8ac5fc53764fc058c3f88404a8

cd1m0 (approved):
LGTM

PR #46, opened 2021-05-24
Cleanup instrumentation backend

cd1m0:
The new addition of `forall` exposed 2 problems with the "design" of the original instrumentation code. Namely:

1. The design was centered on the assumption that transpiling can be split into 2 phases - a) flattening (breaking down a complex expression into a list of assignments and b) translation from Scribble to Solidity.
This assumption is now broken by `forall`-s - which require insertion of (potentially nested) loops with their own scopes, as well as potentially emitting assignments inside the loops as opposed to just at the top-level
2. Splitting re-writing into 2 parts - a) generating a list of rewrite steps (recipe) and b) executing (cooking) the recipe. 
This was originally a neat idea, however as the code grew, I found myself finding shortcuts away from building recipes due to the verbosity of the code. This half-way approach of sometimes using recipes, and sometimes not is confusing.

To fix those 2 problem this PR does the following:

1. Unify the old flattening (`flattenExpr`) and transpiling (`generateExprAst`) passes into a single transpiling (`transpile`) pass.
2. Remove all recipes and replace them with eager re-writes in-place.


blitz-1306:
The other reason that **solc-typed-ast**'s `ASTNodeFactory` and ASTNode's insert/remove logic became a bit more reliable. So it is good decision to remove some obsolete code.

src/instrumenter/transpile.ts
```diff
@@ -93,61 +98,57 @@ export function generateTypeAst(type: TypeNode, factory: ASTNodeFactory): TypeNa
...
+    if (BuiltinSymbols.has(expr.name)) {
+        return factory.makeIdentifier("<missing>", expr.name, -1);
     }
 
-    return DataLocation.Default;
-}
+    // Builtin symbols are the only identifiers with undefined `defSite`.
+    if (expr.defSite === undefined) {
+        throw new Error(
+            `Cannot generate AST for id ${expr.pp()} with no corrseponding definition.`
```
blitz-1306, line 115: Typo in `corresponding`

blitz-1306 (dismissed):
LGTM

src/instrumenter/transpile.ts
cd1m0: Fixed in f9494de

blitz-1306 (approved):

PR #50, opened 2021-05-28
Tests using artefacts

blitz-1306:
## Changes
- [x] Intoduced logic to build custom artefacts for tests (`ExtendedCompileResult`, `removeProcWd()`, `makeArtefact()`, `toAstUsingCache()`).
- [x] Tweaked tests to use artefacts instead of invoking compiler multiple times for same source files.

## Notes
- Artefacts are "custom":
  - Not contains identation and `legacyAST` to occupy less space on drive.
  - Additionally contains `compilerVersion`, that was used to generate artefact.
- Artefacts have to be updated if any related source is changed. There is a validation test `artefacts.spec.ts`, checking that artefacts are valid.

Regards.


cd1m0 (approved):
lgtm

PR #55, opened 2021-07-22
Unchecked sum over maps

cd1m0:
This PR adds unchecked sum over maps, on top of map quantification. It also does several small cleanups for forall over maps.

1. Cleanups for forall over maps
      a) Emit various functions in custom map libraries on demand, to make output more succinct
      b) Add more testing around forall and old, and fix one edge case in scForAll
      
2. Support for `unchecked_sum()` over maps:
     a) Add a `sum` field in the struct for all emitted custom map libraries that have a numeric value type
     b) Modify the setter function in custom map libraries to update the the sum variable
     c) Modify type-checker to handle `unchecked_sum(map)`. Set the type of sum to a 256-bit integer with the same sign as the sign of values in the underlying map
     d) Modify the semantic-checker to add maps appearing inside of `unchecked_sum` to the queue of maps to be interposed
     e) Modify the transpiler to transpile `unchecked_sum(map)` as an access to the underyling sum field
     d) Add integration tests


src/instrumenter/custom_maps_templates.ts
```diff
@@ -698,6 +730,27 @@ function makeSetFun(
     const mkKeysLen = () => factory.makeMemberAccess("<missing>", mkKeys(), "length", -1);
     const mkKeysPush = () => factory.makeMemberAccess("<missing>", mkKeys(), "push", -1);
     const mkKeyIdxM = () => mkStructFieldAcc(factory, factory.makeIdentifierFor(m), struct, 2);
+    const mkSum = () => mkStructFieldAcc(factory, factory.makeIdentifierFor(m), struct, 3);
+
+    if (valueT instanceof IntType) {
+        // TODO: There is risk of overflow/underflow here
```
blitz-1306, line 736: Question: is there a still `TODO` or it is handled by current `if` branch?

blitz-1306 (approved):
LGTM. Just one minor question.

src/instrumenter/custom_maps_templates.ts
cd1m0: Yep there is still risk. For now we are accepting this as a feature, and communicating it to the user by renaming the function to `unchecked_sum`

blitz-1306 (approved):

blitz-1306 (approved):

PR #52, opened 2021-06-25
Support new logging debug args

norhh:
TODO: Add a lot of tests

PR #58, opened 2021-07-27
Support/old debug ids

### LINKED ISSUE #56: --debug-events flag doesn't emit correct value for variables inside of `old()`
What the title says :). If you have an annotation like `old(a+1) == b`, and you add `--debug-events` then the value that would be emitted in the events for `a` would be its new value, not its old value. We should fix this.

cd1m0:
This pr is a follow-up on the new debug events logic (#52 ). It does the following:

1. Closes #56 
2. Refactor some of the logic that was in `gatherDebugIds` and `getDebugInfo` to be placed in `transpileId` and `TranspilingContext` which makes it cleaner
3. Allow for arrays to appear in debug events
4. Add iterator variables in `forall`s to the ids that we emit
5. Add any path ids that appear in `if_assigned` to the values that we emit
6. Add functional tests to check that we get the right behavior in the face of `old()`


src/instrumenter/transpile.ts
```diff
@@ -131,6 +133,99 @@ export function transpileType(type: TypeNode, factory: ASTNodeFactory): TypeName
...
+ * the current `TranspilingContext` `ctx`, if `--debug-events` are on, and this
+ * is the first time we encounter `id`, and its type is traceable (see
+ * `isTypeTraceable`) then add debugging information for `id` to `ctx`.
+ *
+ * Note that special care needs to be taken for ids appearing in old contexts -
+ * we need to add a new binding for them that stores their value.
+ *
+ * @param id
+ * @param transpiledExpr
+ * @param ctx
```
blitz-1306, line 195: Empty params description, that may be removed.

src/instrumenter/transpile.ts
```diff
@@ -406,21 +509,41 @@ function transpileFunctionCall(expr: SFunctionCall, ctx: TranspilingContext): Ex
...
+
+            const struct = single(lib.vStructs);
+            return mkStructFieldAcc(factory, transpile(arg, ctx), struct, "sum");
+        }
+
         assert(
-            sVar !== undefined,
-            `sum argument should be a state var(or a part of it), not ${arg.pp()}`
+            argT.to instanceof ArrayType,
+            `sum expects a pointer to array/map, not ${argT.pp()}`
```
blitz-1306, line 536: A note, that does not unrgently require a fix: `pp()`-ing in `assert()`s is slow, especially if used in message template string. This is because arguments are resolved on each call, so `pp()` is actually executed no matter if assertion is `true` or `false`. I guess we need `assert()` optimization in **solc-typed-ast**, or local optimization to boost performance.

blitz-1306 (commented):
Looks good. One small nit and one sidenote.

src/instrumenter/transpile.ts
cd1m0: Fixed in 3084bb1

src/instrumenter/transpile.ts
cd1m0: Yeah, lets do a separate cleanup pass for that. I have 2-3 things in mind we can do in a single cleanup pass. Lets discuss offline

blitz-1306 (approved):

norhh (approved):
👍

PR #59, opened 2021-08-01
Fix bug in arming code with cutom maps and solidty < 0.8.6

cd1m0:
The recent custom map implementaiton introduces a bug when compiling code <0.8.0. Specifically, the custom maps library template includes an `unchecked()` block regardless of the compiler version, which doesn't compile on solc <0.8.0. Fix is to only emit that blog for the right compiler versions.


blitz-1306 (approved):
Looks good.

PR #70, opened 2021-09-13
Fix crashes on ternary operators on the RHS of tuple assignments

### LINKED ISSUE #62: Scribble fails on ternary operator in tuple assignment
```solidity
function f(bool cond) internal pure returns (uint[2] memory ret) {
  (ret[0], ret[1]) = cond ? (1, 2) : (3, 4);
}
```

results in

```
Error: Unexpected rhs in tuple assignment: Conditional#1710
    at getAssignmentComponents (/Users/alex/.nvm/versions/node/v16.8.0/lib/node_modules/eth-scribble/dist/instrumenter/state_vars.js:41:19)
    at getAssignmentComponents.next (<anonymous>)
    at getAssignments (/Users/alex/.nvm/versions/node/v16.8.0/lib/node_modules/eth-scribble/dist/instrumenter/state_vars.js:124:20)
    at getAssignments.next (<anonymous>)
    at Object.findAliasedStateVars (/Users/alex/.nvm/versions/node/v16.8.0/lib/node_modules/eth-scribble/dist/instrumenter/state_vars.js:272:21)
    at new InstrumentationContext (/Users/alex/.nvm/versions/node/v16.8.0/lib/node_modules/eth-scribble/dist/instrumenter/instrumentation_context.js:231:38)
    at Object.<anonymous> (/Users/alex/.nvm/versions/node/v16.8.0/lib/node_modules/eth-scribble/dist/bin/scribble.js:582:26)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
```

cd1m0:
This PR fixes #62 and depends on #68. To do so we also tweak the way we instrument tuple assignments when one of the RHS expressions is a storage pointer. In the past if we did:

```
   (... , storArr, ....) = (...., anotherStorArr, ...)
```

Where both `storArr` and `anotherStorArr` are arrays in storage, we would introduce a temporary storage pointer for the interposition:

```
   uint[] storage tempStorePtr;
   (... , tempStorePtr, ....) = (...., anotherStorArr, ...)
   ...
   storArr = tempStorePtr; 
```

This required some brittle logic, but saved us an extra expensive copy in the instrumented code in these cases. However that fragile logic breaks when we try to handle conditionals in tuple assignments (see the difference of `replaceLHS` in `state_var_instrumenter.ts` for details). Fixing this requires adding more complexity.

Instead of this, I decided to go with a simpler solution, of just always making the temporaries live in memory, which in some rare cases may add some overhead in the generated code, but results in simpler and more maintainable scribble code.


src/instrumenter/state_vars.ts
```diff
@@ -42,36 +43,59 @@ export type RHS = Expression | [Expression, number];
...
+            assignedDecls = lhs.assignments.map((declId) =>
+                declId === null ? null : (lhs.requiredContext.locate(declId) as VariableDeclaration)
+            );
+        }
+
+        if (assignedDecls.length === 1) {
+            if (assignedDecls[0] !== null) {
+                yield* getAssignmentComponents(assignedDecls[0], rhs);
             }
         } else if (rhs instanceof TupleExpression) {
```
blitz-1306, line 65: I'm wondering if `rhs` of `TupleExpression` with truthy [`isInlineArray`](https://github.com/ConsenSys/solc-typed-ast/blob/64c1d1e3ee4bba27a25ca46a4e8d9fc494b1bbe5/src/ast/implementation/expression/tuple_expression.ts#L8) may cause trouble. Maybe not, because assigned declarations will still match only whole array literal and not it's parts. But if there would be a way to check such edge-case - it would be cool.

blitz-1306 (approved):
Looks good. Only one suspicion, but it may be just me being a paranoid.

src/instrumenter/state_vars.ts
cd1m0: So the check is on the `lhs`, and I don't think a LHS expression can ever be an inline array, which is why I removed it

PR #71, opened 2021-09-13
Fix scribble crash on contracts with modifier invocation call to base contract along with InheritanceSpecifier for same base contract

### LINKED ISSUE #63: Scribble fails on explicit call to inherited constructor.
The following two examples are accepted by Scribble:
```solidity
contract A {
    int public a;
    constructor() {}
}

/// #invariant a >= 0;
contract B is A {
    constructor() {}
}
```
```solidity
contract A {
    int public a;
    constructor(int _a) { require(_a >= 0); a = _a; }
}

/// #invariant a > 0;
contract B is A(1) {
    constructor() {}
}
```
The following example yields `Error: Mismatch in length between [VariableDeclaration#6] of len 1 and [] of len 0`:
```
contract A {
    uint public a;
    constructor(uint _a) { require(_a >= 0); a = _a; }
}

/// #invariant a > 0;
contract B is A {
    constructor() A(1) {}
}
```
I am running the latest version of Scribble on npm using node version `10.24.1`.

cd1m0:
Fixes #63 and depends on #70


blitz-1306 (approved):
Looks good to merge. Still there is some PR dependency that we ferst need to handle properly.

PR #76, opened 2021-09-16
Make debug event signature more robust

cd1m0:
Currently we describe the structure of the encoded bytes field in the instrumentation metadata as a list of pairs of strings - `[variable name, variable type][]`. However there may be multiple variables with the same name, so this doesn't work. A better (and more future-proof) scenario is to report the following structure: `[string[], string]`. Here the first element is a list of source triples, describing locations where a given variable appears in the annotation, and the second element is the type of the data. So for example given the following annotation:

```
/// if_succeeds let x:= x+x in x*x == 16;
function foo(uint x) ...
```

We would return the following description of debug information in the metadata:

```
     [
         [
             "20:1:0", // location correspond to the first x in "x+x"
             "22:1:0" // location correspond to the second x in "x+x"
         ],
         "uint"
    ],
    [
        [
             "16:1:0", // location corresponding to the x in `let x:=`
             "27:1:0", // location corresponding to the first x in `x*x`
             "29:1:0"// location corresponding to the second x in `x*x`
       ],
       "uint"
    ]
```

This format should make it easy for the frontend to know where each value should be reported.


blitz-1306 (approved):
Looks good.

norhh (approved):
👍

PR #84, opened 2021-09-24
Emit a type error message when we cant determine the type of a binding

### LINKED ISSUE #74: Scribble crashes on literals in certain places
The following sample cause a crash:

```
/// #invariant let a := 1 in a > 0;
contract Foo {
}
```
Stack trace:

```
/home/dimo/work/consensys/scribble/dist/bin/scribble.js:484
            throw e;
            ^

Error: Unsupported spec type int_const
    at transpileType (/home/dimo/work/consensys/scribble/dist/instrumenter/transpile.js:21:15)
    at transpileLet (/home/dimo/work/consensys/scribble/dist/instrumenter/transpile.js:387:22)
    at transpile (/home/dimo/work/consensys/scribble/dist/instrumenter/transpile.js:552:16)
    at Object.transpileAnnotation (/home/dimo/work/consensys/scribble/dist/instrumenter/transpile.js:494:16)
    at insertAnnotations (/home/dimo/work/consensys/scribble/dist/instrumenter/instrument.js:225:42)
    at makeInternalInvariantChecker (/home/dimo/work/consensys/scribble/dist/instrumenter/instrument.js:325:5)
    at Object.instrumentContract (/home/dimo/work/consensys/scribble/dist/instrumenter/instrument.js:258:36)
    at instrumentFiles (/home/dimo/work/consensys/scribble/dist/bin/scribble.js:185:17)
    at Object.<anonymous> (/home/dimo/work/consensys/scribble/dist/bin/scribble.js:478:13)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)

```
    
The problem is that we are not inferring a type for the underlying temporary used to store a. For now the user can avoid this by just adding a cast:
    
```
    /// #invariant let a := uint(1) in a > 0;
```
    
We can try to pick the smallest type that fits the literal, but this may cause other problems, for example in this case ```/// invariant  let a:= 1 in a == int(1)```. This would fail type checking with this scheme, as the smallest type to fit the `1` is `uint8` but `a` is used in the body of the let as an `int256`.

Another solution would be to produce a proper error message, that asks the user to qualify the type of the temporary. The same problem can be hit with the following syntactic constructs:

```
/// #invariant (let a := true in 0) == 0;
contract Foo {
}
```

```
contract Foo {
        /// #if_succeeds old(0) > 0;
        function foo() internal {}
}
```

cd1m0:
This PR fixes #74 .

Specifically when we can't determine the type of a let-var, or the entire let-binding we emit an error message asking the user to provide a cast. This usually happens when the RHS expression is a constant numeric expression.

When we encounter a constant numeric expression inside of `old()` we silently ignore the `old()`.


blitz-1306 (approved):
Looks good

cd1m0:
There were some broken tests in this PR that I missed. I ended up fixing them while working on another followup PR - #90 . So closing this in favour of #90  which also fixes this issue.

PR #87, opened 2021-09-30
Fixes for #85 and #86

### LINKED ISSUE #85: Scribble crashes in the presence of assingments from storage pointers to storage pointers
Consider the following sample:

```solidity
contract Foo {
    struct S {
        uint[] arr;
    }

    function main(S storage v) private {
        S storage v1 = v;
    }
}
```

Running scribble (up to version 0.5.5) on this produces the following crash:

```
/home/dimo/work/consensys/scribble-clean/dist/instrumenter/state_vars.js:321
        throw new Error(`Unexpected RHS element ${__1.print(rhs)} in assignment to state var pointer`);
        ^

Error: Unexpected RHS element v in assignment to state var pointer
    at gatherRHSVars (/home/dimo/work/consensys/scribble-clean/dist/instrumenter/state_vars.js:321:15)
    at Object.findAliasedStateVars (/home/dimo/work/consensys/scribble-clean/dist/instrumenter/state_vars.js:350:26)
    at new InstrumentationContext (/home/dimo/work/consensys/scribble-clean/dist/instrumenter/instrumentation_context.js:231:38)
    at Object.<anonymous> (/home/dimo/work/consensys/scribble-clean/dist/bin/scribble.js:487:26)
    at Module._compile (internal/modules/cjs/loader.js:1015:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1035:10)
    at Module.load (internal/modules/cjs/loader.js:879:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12)
    at internal/main/run_main_module.js:17:47
```

The issue is that the logic of this does not account for the case where the RHS of an assignment to a storage pointer is an Identifier, but its not a direct reference to a state var. This happens when the RHS is another storage pointer. (see code below):

```typescript
    /**
     * Given a potentially complex RHS expression, return the list of
     * state variable declarations that it may alias
     */
    const gatherRHSVars = (rhs: Expression): VariableDeclaration[] => {
        if (isStateVarRef(rhs)) { // <-- This check needs to handle all Identifiers
            return [rhs.vReferencedDeclaration as VariableDeclaration];
        }

...
        throw new Error(`Unexpected RHS element ${print(rhs)} in assignment to state var pointer`);
    };
```

### LINKED ISSUE #86: Scribble needs to disallow equality comparisons between reference types
Consider the following sample:

```solidity
contract Foo {
    struct S {
        uint[] arr;
    }
    /// #if_succeeds a == b;
    function main(S memory a, S memory b) private {
    }
}
```

Scrible's type-checker will allow the expression `a==b` and transpile it into the following code:

```solidity
...
    function main(S memory a, S memory b) private {
        _original_Foo_main(a, b);
        unchecked {
            if (!(a == b)) {
                emit AssertionFailed("0: ");
                assert(false);
            }
        }
    }
...
```

That code however doesn't compile since you can't compare pointers in Solidity. To fix this the scribble type-checker should be made more precise, to disallow equality comparison between reference types.

cd1m0:
This PR:
 - Fixes #85 
 - Fixes #86

### DIFF
--- a/src/instrumenter/state_vars.ts
+++ b/src/instrumenter/state_vars.ts
@@ -385,6 +385,10 @@ export function findAliasedStateVars(units: SourceUnit[]): Map<VariableDeclarati
             return [rhs.vReferencedDeclaration as VariableDeclaration];
         }
 
+        if (rhs instanceof Identifier) {
+            return [];
+        }
+
         if (rhs instanceof MemberAccess) {
             return gatherRHSVars(rhs.vExpression);
         }
--- a/src/spec-lang/tc/typecheck.ts
+++ b/src/spec-lang/tc/typecheck.ts
@@ -1226,7 +1226,17 @@ export function tcBinary(expr: SBinaryOperation, ctx: STypingCtx, typeEnv: TypeE
 
     if (["==", "!="].includes(expr.op)) {
         // Equality operators allow for the same or implicitly castable types.
-        unifyTypes(expr.left, lhsT, expr.right, rhsT, expr);
+        const commonType = unifyTypes(expr.left, lhsT, expr.right, rhsT, expr);
+
+        if (commonType instanceof PointerType) {
+            throw new SWrongType(
+                `Operator ${
+                    expr.op
+                } doesn't apply to expression ${expr.left.pp()} of non-value type ${lhsT.pp()}`,
+                expr.left,
+                lhsT
+            );
+        }
         return new BoolType();
     }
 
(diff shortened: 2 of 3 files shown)


blitz-1306 (approved):

blitz-1306:
Strange - referenced issues were not closed on PR merge. Maybe a bulleted list causes this.

Issue #88, opened 2021-09-30
Scribble crashes when mixing `for_all` and `if_updated` on nested mappings


cd1m0:
Consider the following sample:
```
contract Foo {
    /// #if_updated forall(uint k in m) m[k][0] > 1;
    mapping(uint=>mapping(uint => uint)) m;

    function main() public {
        m[0][1] = 1;
    }
}```

Running scribble 0.5.5 on this sample produces the following crash:

```
/home/dimo/work/consensys/scribble-clean/dist/bin/scribble.js:500
            throw e;
            ^

Error
    at Object.assert (/home/dimo/work/consensys/scribble-clean/dist/util/misc.js:16:11)
    at decomposeLHS (/home/dimo/work/consensys/scribble-clean/dist/instrumenter/state_vars.js:395:9)
    at addStateVarUpdateLocDesc (/home/dimo/work/consensys/scribble-clean/dist/instrumenter/state_vars.js:456:33)
    at Object.findStateVarUpdates (/home/dimo/work/consensys/scribble-clean/dist/instrumenter/state_vars.js:481:13)
    at instrumentFiles (/home/dimo/work/consensys/scribble-clean/dist/bin/scribble.js:204:37)
    at Object.<anonymous> (/home/dimo/work/consensys/scribble-clean/dist/bin/scribble.js:494:13)
    at Module._compile (internal/modules/cjs/loader.js:1015:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1035:10)
    at Module.load (internal/modules/cjs/loader.js:879:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
```

The issue is that the logic inside `decomposeLHS` is not written to handle functions returning a storage pointer, since scribble doesn't allow instrumenting state variables that are aliased. However the custom mapping instrumentation due to the `forall` introduces a call to `<custom map library>.get_lhs(...)` on the LHS of an assignment. To fix this we need to make `decomposeLHS` aware of potential changes done by the custom map instrumentation.

cd1m0:
Fixed with #90

PR #90, opened 2021-09-30
Fixes for #74 and #88

cd1m0:
This PR provides fixes for:

- #74 
- #88 

Specifically for #74 we emit a type error message whenever the user wrote an annotation, with literals that cause ambiguities in the underlying instrumentation.

For #88 we update the state var aliasing logic to handle cases where the LHS of an assignment has been instrumented into a `get_lhs` call due to custom map instrumentation.


blitz-1306 (commented):
One minor nit, otherwise looks good.

test/integration/artefacts.spec.ts
```diff
@@ -23,7 +23,7 @@ describe("Artefacts validation", () => {
             const expected = makeArtefact(toAst(sample));
 
             // Uncomment next line to update all of the artefacts
-            // fse.writeFileSync(artefact, expected);
+            fse.writeFileSync(artefact, expected);
```
blitz-1306, line 26: Should be commented out.

cd1m0: Fixed in 59bbc3d .

blitz-1306 (approved):

PR #94, opened 2021-10-11
Optimize `assert()` calls

blitz-1306:
## Premise
Some time ago **solc-typed-ast** introduced pretty printing routines in https://github.com/ConsenSys/solc-typed-ast/pull/71. This PR removed duplicated codebase and reuses **solc-typed-ast**'s functionality. Also it boosts speed as `pp()` and `ASTNode.print()` deferred until there is a necessity to throw and error and compose a message for it (currently a lot of `pp()`-ing is done, but results are not used).

## Changes
- [x] Updated dependencies.
- [x] Tweak canonical signature composition calls to use newer API: default to ABI encoder v2.
- [x] Removed `src/util/pp.ts` and `src/util/struct_equality.ts` and reused logic from **solc-typed-ast**.
- [x] Cleanup in `assert()` calls to defer `pp()` and `ASTNode.print()` until error is actually thrown.
- [x] Fixed a ton of places when `assert()` call had an empty message.
- [x] Reduced use of `big-integer` library to just support scientific notation in expression parser.
- [x] Slight code readability improvements.
- [x] Tweaked affected test suites.

## Notes
- Were unable to completely get rid of [`big-integer`](https://github.com/peterolson/BigInteger.js) dependency due to currently Scribble supports numbers with scientific notation in predicate expressions. Maybe there is a way to replace it with something light-weight (consider as a followup).

Regards.


src/instrumenter/cha.ts
```diff
@@ -179,16 +176,16 @@ export function getCHA(srcs: SourceUnit[]): CHA<ContractDefinition> {
     }
 
     const res = { parents, children, roots, leaves };
-    assert(checkCHA(res), `Built CHA ${res} is malformed`);
+
+    assert(checkCHA(res), "Built CHA {0} is malformed", ppCHA(res));
```
cd1m0, line 180: `ppCHA` is probably fairly expensive. Maybe move this inside an `if (!checkCHA(res)) { assert(false, "...", ppCHA(res)); }?

src/instrumenter/cha.ts
```diff
@@ -197,13 +194,18 @@ export function chaDFS<T>(cha: CHA<T>, cb: (node: T) => void): void {
...
         cb(cur);
 
         const children = cha.children.get(cur);
+
         assert(
             children !== undefined,
-            `Node ${pp(cur)} is missing from children map in ${ppCHA(cha)}`
+            "Node {0} is missing from children map in {1}",
+            cur,
+            ppCHA(cha)
```
cd1m0, line 206: Same here - `ppCHA` is fairly expensive. Maybe move assert inside of if?

src/spec-lang/tc/typecheck.ts
```diff
@@ -84,7 +87,12 @@ export class StateVarScope {
         public readonly target: VariableDeclaration,
```
cd1m0, line 87: If we add an `implements PPAble` to the class definition for `StateVarScope` can we avoid the casts to `scope as PPAble` further down in this file?

test/integration/src2srcmap.spec.ts
```diff
@@ -173,13 +172,11 @@ describe("Src2src map test", () => {
                     if (instrNodes === undefined) {
```
cd1m0, line 172: We can potentially unnset the if and have a top-level `assert(instrNodes !== undefined, ...)`

cd1m0 (changes requested):
AWESOME work. This makes the code a lot better. Just 3-4 nits - might as well fix them as well while we are doing this

src/instrumenter/cha.ts
blitz-1306: It would be easier to `throw` than `assert()` in such case. Will fix.

blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/94/commits/d4ed33eadbc26a96e9b7719cff33f967783af572

src/instrumenter/cha.ts
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/94/commits/d4ed33eadbc26a96e9b7719cff33f967783af572

src/spec-lang/tc/typecheck.ts
blitz-1306: Good catch. Addressed in https://github.com/ConsenSys/scribble/pull/94/commits/d4ed33eadbc26a96e9b7719cff33f967783af572.

test/integration/src2srcmap.spec.ts
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/94/commits/d4ed33eadbc26a96e9b7719cff33f967783af572

cd1m0 (approved):
lgtm!

cd1m0:
Only CI failures are the intermittent segfaults due to nyc. Force-merging

PR #96, opened 2021-10-21
Support universal annotations

blitz-1306:
##  Preface
This PR introduces logic which consumes a YAML configuration that describes macros and re-write rules to automatically add properties to a set of smart-contracts.

## Status
**WIP**, do not merge. Solution is in R&D / PoC stage.

## Changes
- [x] Introduced macro definitions in YAML.
- [x] Tweaked grammar to support macro calls.
- [x] Added logic to match macro instructions with macro definitions.
- [x] Introduced macro property annotations parsing "on fly" and variable name substitution (with names that are passed in macro instruction), also produced annotation injection.
- [x] Introduced `--macro-path` CLI parameter.
- [x] Tuned `searchRecursive()` and introduced to whole code base (not just tests). Adjusted calls to use newer version of `searchRecursive()`.

Regards.


src/instrumenter/annotations.ts
```diff
@@ -546,6 +608,163 @@ export function gatherContractAnnotations(
...
+                 * @todo Using `inclusive = true` here seems to be a design oversight.
+                 * Better to fix this in **solc-typed-ast**.
+                 */
+                const targets = resolveAny(name, scope, compilerVersion, true);
+                const target = [...targets][0];
+
+                if (target instanceof FunctionDefinition && args.length > 0) {
+                    const params = target.vParameters.vParameters;
+
+                    mapArgsToParams(signature, args, params, localAliases);
```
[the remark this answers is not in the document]
cd1m0: Or even:

```
for (const [formalParam, actualParam] of zip(args, params).filter(([arg,])=>arg !== "")) {
   localAliases[formalParam] = actualParam;
}
```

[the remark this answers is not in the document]
cd1m0: I do realize that in `mapArgsToParams` you have a more specific error message. To preserve that error message, we can add an optional error message argument to `zip()`, similarly to how we have such an argument to `single()`

src/instrumenter/annotations.ts
```diff
@@ -546,6 +608,163 @@ export function gatherContractAnnotations(
...
+        } else {
+            metas.push(...metasToInject);
+        }
+    }
+
+    /**
+     * @todo list:
+     * [TODO]   1. Source locations are needed for property maps.
+     * [TODO]   2. Figure out how to use macro variable types.
+     * [TODO]   3. Figure out if macros can use other macros.
```
cd1m0, line 764: As a first iteration lets not allow this. I don't think its straight forward what that would look like. One way to potentially reason about this is to:

1. Limit macros to contract-level annotations only
2. Add "inheritance" amongst macros. Then for example you could have a hierarchy of `ERC20` macros, with increasingly more properties.

One potential issue with this approach is that the children inherit all the properties of the parents. I suspect there may be cases when the "inheriting" macro needs to invalidate properties of the parent. This needs more discussion, so for now lets leave it off the table.

src/instrumenter/annotations.ts
```diff
@@ -546,6 +608,163 @@ export function gatherContractAnnotations(
...
+            annotationMap.set(target, metasToInject);
+        } else {
+            metas.push(...metasToInject);
+        }
+    }
+
+    /**
+     * @todo list:
+     * [TODO]   1. Source locations are needed for property maps.
+     * [TODO]   2. Figure out how to use macro variable types.
```
cd1m0, line 763: What do you mean by variable types?

src/instrumenter/annotations.ts
```diff
@@ -546,6 +608,163 @@ export function gatherContractAnnotations(
...
+        if (metas === undefined) {
+            annotationMap.set(target, metasToInject);
+        } else {
+            metas.push(...metasToInject);
+        }
+    }
+
+    /**
+     * @todo list:
+     * [TODO]   1. Source locations are needed for property maps.
```
cd1m0, line 762: So this is a really tricky problem. Whereas ast nodes in normal annotations have a single source location (wherever the original annotation in the docstring is), ast nodes in instantiated macros actually have **TWO** locations:

1. The source location corresponding to the AST node inside the annotation in the yaml file
2. The location of the actual macro invocation on the target contract.

And when we want to report errors, ideally we want to show the users something like:

Error on line X col Y in file macros.yaml: Error unknown identifier 'x' in /// invaraiant x>0;
Instantiated on line Z col W in file target.sol: /// #ERC20( .... )

To support this cleanly we could do something like change the type of `src` in `SNode` to something weird like `string | MacroSrc` where `MacroSrc` contains both sources. However this is probably going to be a very invasive change.

src/spec-lang/expr_grammar.pegjs
```diff
@@ -11,7 +11,8 @@ Annotation =
         / UserFunctionDefinition
```
cd1m0, line 11: In the Annotation rule, shouldn't the `prefix` now be reqiured? I.e. it should be:

```
          prefix: '#'
```

src/instrumenter/annotations.ts
```diff
@@ -267,242 +316,255 @@ type RawMetaData = {
...
+    macros: Map<string, MacroDefinition>
+): AnnotationMetaData[] {
+    const rxType = filters.type === undefined ? undefined : new RegExp(filters.type);
+    const rxMsg = filters.message === undefined ? undefined : new RegExp(filters.message);
 
-        const result: AnnotationMetaData[] = [];
+    const sourceInfo = raw.sourceInfo;
 
-        const rx =
-            /\s*(\*|\/\/\/)\s*#?(if_succeeds|if_updated|if_assigned|invariant|assert|try|require|define\s*[a-zA-Z0-9_]*\s*\([^)]*\))/g;
```
cd1m0, line 453: Now that '#' is required, I think the `#` in this regex be required?

cd1m0 (changes requested):
Awesome work @blitz-1306 ! There is work left, but this is a really good start. There are some comments for code style issues, and there are some comments about general discussion about how to deal with the remaining todos, and other more general design issues. If you can address the style issues, we can discuss in parallel how to resolve the design issues outlined

src/instrumenter/annotations.ts
```diff
@@ -267,242 +316,255 @@ type RawMetaData = {
...
+        loc: [sourceInfo.offset, sourceInfo.length]
+    };
 
-        let match = rx.exec(meta.text);
+    const result: AnnotationMetaData[] = [];
 
-        while (match !== null) {
-            const annotation = this.makeAnnotationFromMatch(match, meta, source);
+    const rx =
+        /\s*(\*|\/\/\/)\s*#(if_succeeds|if_updated|if_assigned|invariant|assert|try|require|define\s*[a-zA-Z0-9_]*\s*\([^)]*\)|[a-zA-Z0-9_]*\s*\([^)]*\))/g;
```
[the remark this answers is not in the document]
blitz-1306: Yeah, that was my second option. Lets do that.

src/instrumenter/annotations.ts
blitz-1306: Macro variables can be defined with the types, however those types are never user right now. There is no typechecking that macro call values are matching macro variable types.

src/instrumenter/annotations.ts
```diff
@@ -546,6 +608,163 @@ export function gatherContractAnnotations(
...
+                        }
+                    }
+                };
+
+                /**
+                 * @todo Using `inclusive = true` here seems to be a design oversight.
+                 * Better to fix this in **solc-typed-ast**.
+                 */
+                const targets = resolveAny(name, scope, compilerVersion, true);
+                const target = [...targets][0];
```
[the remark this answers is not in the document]
blitz-1306: There may be overloads and overrides, I guess here we should pick first accessible instance and `throw` if there is nothing to pick.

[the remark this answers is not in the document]
cd1m0: If there is overriding, `resolveAny` should still return a single function. Overloading however is a problem. If there are multiple overloaded functions, then we have two choices:

1. Throw an error saying this case is not yet supported
2. Use the parsed `args` to resolve the singualr correct overloaded function.

src/instrumenter/annotations.ts
blitz-1306: `#` is no longer required due to https://github.com/ConsenSys/scribble/pull/96/commits/e57764f3acdbedbb98e4e3d041446f2210f42531 introduces `macro` keyword/

src/spec-lang/expr_grammar.pegjs
blitz-1306: `#` is no longer required due to https://github.com/ConsenSys/scribble/pull/96/commits/e57764f3acdbedbb98e4e3d041446f2210f42531 introduces `macro` keyword.

src/instrumenter/annotations.ts
blitz-1306: Deleted point 3 in https://github.com/ConsenSys/scribble/pull/96/commits/e57764f3acdbedbb98e4e3d041446f2210f42531. Currently only `SProperty` descendants are allowed (and `SMacro` is not one of them).

src/instrumenter/annotations.ts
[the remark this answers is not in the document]
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/96/commits/e57764f3acdbedbb98e4e3d041446f2210f42531

src/instrumenter/annotations.ts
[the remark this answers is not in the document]
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/96/commits/e57764f3acdbedbb98e4e3d041446f2210f42531

src/instrumenter/annotations.ts
```diff
@@ -546,6 +608,163 @@ export function gatherContractAnnotations(
...
+                        const actualName = localAliases.get(node.name);
+
+                        if (actualName !== undefined) {
+                            node.name = actualName;
+                        }
+                    }
+                };
+
+                /**
+                 * @todo Using `inclusive = true` here seems to be a design oversight.
```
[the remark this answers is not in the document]
blitz-1306: Removed `@todo` in https://github.com/ConsenSys/scribble/pull/96/commits/e57764f3acdbedbb98e4e3d041446f2210f42531. Just got confused with it.

src/instrumenter/annotations.ts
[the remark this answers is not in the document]
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/96/commits/e57764f3acdbedbb98e4e3d041446f2210f42531

src/instrumenter/annotations.ts
```diff
@@ -237,6 +246,46 @@ export class PropertyMetaData extends AnnotationMetaData<SProperty> {
...
+
+        this.macroDefinition = macroDefinition;
+    }
+
+    /**
+     * Computes a mapping where keys are formal variable names from macro definition
+     * and values are actual names from macro instruction on contract.
+     */
+    getAliasingMap(): Map<string, string> {
+        const result = new Map<string, string>();
```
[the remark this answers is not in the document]
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/96/commits/e57764f3acdbedbb98e4e3d041446f2210f42531

src/instrumenter/annotations.ts
blitz-1306: If we introduce macros as an **experimental feature**, we can try to map produced annotations to macro annotation that is present in the source. It would be very broad, still it will not block merging. We should have a disclaimer that should explicitly state that "_implementation is still not in final state_".

src/instrumenter/annotations.ts
```diff
@@ -267,242 +316,255 @@ type RawMetaData = {
     loc: OffsetRange;
 };
 
-class AnnotationExtractor {
```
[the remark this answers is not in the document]
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/96/commits/3e2b8729eb9b00f87328f690a1fb5e06b2fe445c

src/instrumenter/annotations.ts
```diff
@@ -676,36 +659,81 @@ function processMacroAnnotations(annotationMap: AnnotationMap, compilerVersion:
...
+                    name,
+                    scope
+                );
 
                 if (target instanceof FunctionDefinition && args.length > 0) {
                     const params = target.vParameters.vParameters;
+                    const pairs = zip(
+                        args,
+                        params,
+                        `{0}: arguments count {1} in macro definition exceeds method arguments count {2}`,
```
cd1m0, line 675: The error message here doesn't match the behavior of `zip`. `zip` will fail if the lengths are not exactly equal.

cd1m0 (commented):
Awesome work! Just one nit. I will take a stab at the TODOs left in `processMacroAnnotaitons`.
Also I kind of want to vote to disallow `VariableDeclaration`s as macro targets to limit the scope and complexity of this PR.  @blitz-1306  what do you think?

src/instrumenter/annotations.ts
```diff
@@ -267,242 +319,253 @@ type RawMetaData = {
...
+        loc: [sourceInfo.offset, sourceInfo.length]
+    };
 
-        let match = rx.exec(meta.text);
+    const result: AnnotationMetaData[] = [];
 
-        while (match !== null) {
-            const annotation = this.makeAnnotationFromMatch(match, meta, source);
+    const rx =
+        /\s*(\*|\/\/\/)\s*#?(if_succeeds|if_updated|if_assigned|invariant|assert|try|require|define|macro\s*[a-zA-Z0-9_]*\s*\([^)]*\))/g;
```
cd1m0, line 519: Nit: The modification of the pattern lost the requirement to have a `word(...)` after `define`

src/instrumenter/annotations.ts
cd1m0: So I am thinking lets go the route of C macros - basically we don't check the types of variables at macro instantiation, and instead we just type check the inlined invariants. This may result in weird type errors for now.

However if we fix the node sources to keep track which macro definition an annotation came from, then when we encounter type errors we will correctly report that the type error was in the body of annotation X instantiated in location Y as part of macro Z. This should be enough information for now for users to make progress.

src/instrumenter/annotations.ts
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/96/commits/24012c080b685c49cbcc930f86a107cfee582a84

src/instrumenter/annotations.ts
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/96/commits/24012c080b685c49cbcc930f86a107cfee582a84

blitz-1306:
@cd1m0 I removed support of use case with using macros on state variables. We can always introduce it when there would be an agreement on how they would work really.

src/instrumenter/annotations.ts
```diff
@@ -267,242 +316,255 @@ type RawMetaData = {
...
+        }
+    } else {
+        if (
+            annotation.type !== AnnotationType.IfUpdated &&
+            annotation.type !== AnnotationType.IfAssigned &&
+            /**
+             * @todo Figure out handling the difference between syntax sugar macros for state vars
+             * and contract level macros. This is not yet clear.
+             */
+            annotation.type !== AnnotationType.Macro
```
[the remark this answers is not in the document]
blitz-1306: This is removed in https://github.com/ConsenSys/scribble/pull/96/commits/24012c080b685c49cbcc930f86a107cfee582a84

src/instrumenter/annotations.ts
cd1m0: Fixed with 072f378

src/instrumenter/annotations.ts
cd1m0: Fixed with 072f378

cd1m0 (approved):
LGTM

cd1m0 (approved):
lgtm

PR #97, opened 2021-10-25
Support custom NatSpec tag

### LINKED ISSUE #67: Support custom NatSpec tags
Since Solidity 0.8.2 (March 2021) custom NatSpec tags are supported (see the [docs](https://docs.soliditylang.org/en/latest/natspec-format.html?highlight=custom#tags) and [this discussion](https://github.com/ethereum/solidity/issues/8802) for context).

It would be nice for Scribble to support it, e.g.:
```
/// @custom:scribble if_succeeds {:msg "P0"} y == x + 1;
```

I understand that for supporting older language versions the current syntax should also be kept supported, but at least users' of newer versions can move on to custom tags, which means their NatSpec output will not have Scribble instructions in their user facing `@notice` tags.

norhh:
Supports `@custom:scribble` tag.
closes #67

### DIFF
--- a/src/instrumenter/annotations.ts
+++ b/src/instrumenter/annotations.ts
@@ -366,7 +366,7 @@ function findAnnotations(
     const result: AnnotationMetaData[] = [];
 
     const rx =
-        /\s*(\*|\/\/\/)\s*#?(if_succeeds|if_updated|if_assigned|invariant|assert|try|require|((macro|define)\s*[a-zA-Z0-9_]*\s*\([^)]*\)))/g;
+        /\s*(\*|\/\/\/)\s*(@custom:scribble)?\s*#?(if_succeeds|if_updated|if_assigned|invariant|assert|try|require|((macro|define)\s*[a-zA-Z0-9_]*\s*\([^)]*\)))/g;
 
     let match = rx.exec(meta.text);
 
--- a/test/samples/arr_sum_custom_tag.sol
+++ b/test/samples/arr_sum_custom_tag.sol
@@ -1,0 +1,40 @@
+pragma solidity 0.8.4;
+
+contract Foo {
+    // Simple map
+    uint[] a = [10];
+    int8[] b;
+
+    /// checks for unchecked_sum() @custom:scribble #if_succeeds unchecked_sum(a) > 10 && unchecked_sum(a) < 20;
+    function pushA(uint k) public {
+        a.push(k);
+    }
+
+    /// checks for unchecked_sum() @custom:scribble #if_succeeds unchecked_sum(a) > 10 && unchecked_sum(a) < 20;
+    function setA(uint k, uint v) public {
+        a[k] = v;
+    }
+
+    /// checks for unchecked_sum() @custom:scribble #if_succeeds unchecked_sum(b) > -10 && unchecked_sum(b) < 10;
+    function pushB(int8 k) public {
+        b.push(k);
+    }
+
+    /// checks for unchecked_sum() @custom:scribble #if_succeeds unchecked_sum(b) > -10 && unchecked_sum(b) < 10;
+    function setB(uint k, int8 v) public {
+        b[k] = v;
+    }
+    
+
+    /// checks for unchecked_sum() @custom:scribble #if_succeeds unchecked_sum(c) > -10 && unchecked_sum(c) < 10;
+    function memArr(int16[] memory c) public {
+    }
+
+    /// checks for unchecked_sum() @custom:scribble #if_succeeds unchecked_sum(c) > -10 && unchecked_sum(c) < 10;
+    function calldataArr(int16[] calldata c) external {
+    }
+
+    /// #if_succeeds unchecked_sum(c) < 10;
+    function overflowCheck(uint[] calldata c) external {
+    }
+}
--- a/test/samples/arr_sum_custom_tag.sol
+++ b/test/samples/arr_sum_custom_tag.sol
@@ -5,32 +5,32 @@ contract Foo {
     uint[] a = [10];
     int8[] b;
 
-    /// checks for unchecked_sum() @custom:scribble #if_succeeds unchecked_sum(a) > 10 && unchecked_sum(a) < 20;
+    /// @custom:scribble #if_succeeds unchecked_sum(a) > 10 && unchecked_sum(a) < 20;
     function pushA(uint k) public {
         a.push(k);
     }
 
-    /// checks for unchecked_sum() @custom:scribble #if_succeeds unchecked_sum(a) > 10 && unchecked_sum(a) < 20;
+    /// ignored @custom:scribble #if_succeeds unchecked_sum(a) > 10 && unchecked_sum(a) < 20;
     function setA(uint k, uint v) public {
         a[k] = v;
     }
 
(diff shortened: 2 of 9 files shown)


blitz-1306 (dismissed):
Good job

test/samples/arr_sum_custom_tag.sol
```diff
@@ -0,0 +1,40 @@
...
+    function setA(uint k, uint v) public {
+        a[k] = v;
+    }
+
+    /// checks for unchecked_sum() @custom:scribble #if_succeeds unchecked_sum(b) > -10 && unchecked_sum(b) < 10;
+    function pushB(int8 k) public {
+        b.push(k);
+    }
+
+    /// checks for unchecked_sum() @custom:scribble #if_succeeds unchecked_sum(b) > -10 && unchecked_sum(b) < 10;
```
axic, line 23: Hmm, these shouldn't parse as the `@custom:scribble` tag. Are these being parsed as that?

src/instrumenter/annotations.ts
```diff
@@ -440,10 +439,9 @@ class AnnotationExtractor {
         const result: AnnotationMetaData[] = [];
 
         const rx =
-            /\s*(\*|\/\/\/)\s*#?(if_succeeds|if_updated|if_assigned|invariant|assert|define\s*[a-zA-Z0-9_]*\s*\([^)]*\))/g;
+            /\s*(\*|\/\/\/)\s*((.*)@custom:scribble\s*)?#?(if_succeeds|if_updated|if_assigned|invariant|assert|define\s*[a-zA-Z0-9_]*\s*\([^)]*\))/g;
```
axic, line 442: I think the goal of the custom tag would be to utilise natspec to provide it, as opposed to using regular expressions.

However if you keep using regular expressions, it shouldn't allow leading non-whitespace for the custom tag.

cd1m0: Thanks for reviewing @axic !

I haven't looked at the PR yet - will let you know early next week (sorry working on another project these past 2-3 weeks). From a cursory glance, I think our current implementation just extends our internal regex-y logic to also parse out `@custom` natspec tags instead of relying on the compiler, which I agree is counter-productive to some degree. 

I think the reason we went with this approach for now, was that in order to provide accurate source line information for various errors in the annotation (syntax errors, type errors, etc..) we need to know the relative offset of the start of the annotation to the beginning of the file. If we do it with regex-es ourselves we know exactly where it is in the file.

As far as I could tell, in the parsed natspec information provided in the compiler we couldn't find any indication what was the offset of the annotation relative to the begining of the file (or even the comment). Do you have any ideas how to overcome this?

axic: > I think the reason we went with this approach for now, was that in order to provide accurate source line information for various errors in the annotation (syntax errors, type errors, etc..) we need to know the relative offset of the start of the annotation to the beginning of the file. If we do it with regex-es ourselves we know exactly where it is in the file.
>
> As far as I could tell, in the parsed natspec information provided in the compiler we couldn't find any indication what was the offset of the annotation relative to the begining of the file (or even the comment). Do you have any ideas how to overcome this?

Hmm, this is a good question.  I think the AST view should give you proper source location next to the natspec tags.  (Not only the natspec output, but the AST also contains it under the `documentation` tag.)

Here's an example: https://github.com/ethereum/solidity/blob/develop/test/libsolidity/ASTJSON/documentation.sol and https://github.com/ethereum/solidity/blob/develop/test/libsolidity/ASTJSON/documentation.json

> From a cursory glance, I think our current implementation just extends our internal regex-y logic to also parse out @custom natspec tags instead of relying on the compiler, which I agree is counter-productive to some degree.

If implementing support for the AST would take a longer time, I'd still suggest to strengthen the regex to match the behaviour of upstream, that is not allowing non-whitespace in front of the tag.

axic: Why do you need the location relative in the documentation?

Can't you have two code paths, one for natspec based, and the second for your regex based one?

cd1m0: We could have two paths. I think the problem is it would be bad user experience if we provide accurate source-line information for errors in old-style annotations, but not for natspec annotations.

axic: It should be accurate too with the AST. It tells you the exact position (column/line, but you need to translate it).

Can you please explain what exactly is the issue?

cd1m0: Say we have this code:

```
/// blah blah blah
/// blah
/// @custom:scribble if_succeeds x + y > 0;
function foo(uint x, int y) {
...
}
```

In the ast we know that the whole docstring start at say line 5, col 3. This means that the first 'blah' is around line 5 col 3. However we don't know how much further down the `@custom:scribble` annotation starts. When we parse the contents of the annotation itself, and build an AST we know that the "x+y" starts around 29 characters from the "@". 

Now there is a type error at the "x+y". We know that this is 29 charcters relative to the "@", and we know that the "@" is somewhere inside the docstring that starts at line 5 col 3, but we don't know *exactly* where inside that doc string. For this reason we can't easily convert the error offset "29 characters from the @" into the global error offset "X characters from the start of the file"

cd1m0: I guess we can do an additional pass over each doc string, looking up the starting index of each of the natspec strings inside. Since we know the full text of each natspec string we should be able to disambiguate. The only edge case is if a user enters _exactly_ the same natspec string twice, however we can just make that an error.

blitz-1306: We also support legacy Solidity versions, that do not have `@cusom:whatever`. While we are required to do this, we are unable to change approach. Also we have logic that rebuild `StructuredDocumentation` from source, for 0.4 version that do not have this node in AST. This is an additional historical point to highlight,

test/samples/arr_sum_custom_tag.sol
norhh: These get parsed in `@custom:scribble` when I used --devdoc.

src/instrumenter/annotations.ts
axic: Btw just checked upstream and I think it may be bug that it allows `blabla @cusom:tag test`, because it ignores `blabla` entirely. See https://github.com/ethereum/solidity/issues/12245.

This may be also the reason the offsets may be wrong.

test/samples/arr_sum_custom_tag.sol
cd1m0: I changed the behavior to ignore any lines that have text before `@custom:scribble` on the same line(05abb16). In a followup PR I will add some extra code to issue warnings when something that looks like an invariant has text before that.

src/instrumenter/annotations.ts
cd1m0: For now we are using our old regex/parser approach, as:

1) We need to use it anyways to keep backwards compatibility with older scribble behavior
2) It gives us precise error locations relative to the start of the file (supporting this with natspec will require adding more code)

We changed the behavior to disallow non-white space before the `@custom:scribble`

cd1m0 (approved):
lgtm

blitz-1306 (approved):

cd1m0:
The CI failure is our resident heisen-bug segfault due to nyc mocha. Force-merging.

PR #111, opened 2021-11-16
Fix for #93 - renaming hints and limit annotations to try and require

### LINKED ISSUE #93: Support require-annotations to guide tools
It could be useful to support require-annotation for guiding tools without changing the code itself. This could be very similar to the existing assert-annotations. Below is a more concrete example where some known tokens should be used as inputs:

```Solidity
contract C {
    function foo(address token) public {
      /// #require token == address(0x0) || token == address(0x1234) || token == address(0xabcd);
      ...
      token.transferFrom(...);
```

cd1m0:
This PR Closes #93 . Most of the implementation was done in earlier PR. The goal if this PR is just to:

1. Rename `#hint` and `#limit` to `#try` and `#require`
2. Fix a couple of nits around validating the new annotations in `validateAnnotations`
3. Add a new `type` field to the property metadata so we can distinguish actual property annotations from the new helper annotations


src/spec-lang/tc/semcheck.ts
```diff
@@ -114,7 +114,7 @@ export function scAnnotation(
 ): void {
     if (node instanceof SProperty) {
         // #limit and #hint are implicitly evaluated in the old-state
```
blitz-1306, line 116: Comment still mentioning older `limit` and `hint` keywords. Not a blocker.

blitz-1306 (approved):
LGTM. Small doc issue that is not a blocker for merging.

src/spec-lang/tc/semcheck.ts
cd1m0: Fixed in a1b9613

PR #112, opened 2021-11-16
Add #try, #require and multiple bug fixes

### LINKED ISSUE #93: Support require-annotations to guide tools

### LINKED ISSUE #99: Don't apply contract-level `if_succeeds` annotations to view functions
Currently we support contract-level `if_succeed` annotations that get applied to all non-pure public/external functions of the contract. There is really no need to apply them to the view functions, as they don't change state. However adding these annotations to view functions changes their mutability, which can sometimes cause not to compile. So lets not do it.

### LINKED ISSUE #100: When interposing on constructors super constructor invocations get put on the wrong function.
Given the following code:

```solidity
contract Base {
    constructor(uint x) public {}
}

/// #if_succeeds true;
contract Child is Base {
    constructor() Base(1) {
    }
}
```

When we instrument the `Base(1)` super-constructor invocation gets mistakenly placed on the inner `_original_Child_constructor` function, instead of the constructor:

```solidity
/// #if_succeeds true;
contract Child is Base {

    constructor() {
        _original_Child_constructor();
        unchecked {
            if (!(true)) {
                emit AssertionFailed("0: ");
                assert(false);
            }
        }
    }

    function _original_Child_constructor() Base(1)  private {}
}
```

Thus the resulting code doesn't compile. It should be placed on the constructor.

### LINKED ISSUE #101: Type-checker needs to support library-level constants
The below code will throw a type error at `Foo.Life`. We need to support library constants in the type-checker.

```
library Foo {
        uint256 constant Life = 42;
}

contract Boo {
        /// #if_succeeds $result == Foo.Life;
        function answerOfLife() public returns (uint) {}
}
```

### LINKED ISSUE #103: Broken instrumentation for overriding public variables
Instrumenting the following code:

```
abstract contract Base {
    function a(uint x) external view virtual returns (uint);
}

/// #invariant forall (uint k in a) a[k] > 1;
contract Child is Base{
    mapping(uint=> uint) public override a;
}
```

Produces the following *non-compiling* code:

```
/// #invariant forall (uint k in a) a[k] > 1;
contract Child is __scribble_ReentrancyUtils, Base {
...
    uint256_to_uint256.S override a0;
...
    function a(uint256 ARG_0) public returns (uint256 RET_0) {
        vars2 memory _v;
        _v.__scribble_check_invs_at_end = __scribble_out_of_contract;
        RET_0 = _original_Child_a(ARG_0);
        if (_v.__scribble_check_invs_at_end) __scribble_check_state_invariants();
        __scribble_out_of_contract = _v.__scribble_check_invs_at_end;
    }
```

There are 3 issues here:

1. The `override` keyword should be moved from the state variable `a0` to the introduced getter `a`
2. The getter `a` should have `view` mutability
3. The getter `a` shouldn't be instrumented with checks for state vars since its a view function.

### LINKED ISSUE #104: `forall` broken in new #limit annotation
Given the following sample:

```
contract Foo {
        /// #limit forall (uint k in a) a[k] > 1;
    function foo(uint[] memory a) public {}
}
```

Scribble produces the following broken instrumentation code:

```
pragma solidity 0.7.6;

contract Foo {
...
    function foo(uint[] memory a) public {
        vars0 memory _v;
        require(_v.forall_0);
        _original_Foo_foo(a);
        _v.forall_0 = true;
        for (_v.k0 = 0; _v.k0 < a.length; _v.k0++) {
            _v.forall_0 = a[_v.k0] > 1;
            if (!_v.forall_0) break;
        }
    }

    function _original_Foo_foo(uint[] memory a) private {}
}
```

The issue is that the `require(_v.forall_0);` is placed before the call to the original function, while the computation for the forall is placed after the call.

`limit` is intended to always be evaluated in the `old()` state of a function, so the `forall` should be placed before the function.

Note: As a separate issue we should revisit if the semantics of `limit` (to be implicitly in the `old()` state) makes sense.

### LINKED ISSUE #106: Add optional assertions for checking property coverage
Users want to know what assertions have been covered. This can usually be checked by looking at the coverage data provided by the underlying tool, and mapping it back to source code.

However sometimes this may not be possible (e.g. missing source code, or missing source mappings). For those case, as a quick and dirty workaround, for every actual instrumented property P, we can add "dummy" property P' which fires if `P` is reached. For example for this code:

```
contract Foo {
    /// #if_succeeds true;
    function foo(uint[] memory a) public {}
}
```

We would generate the following code:

```
contract Foo {

    function foo(uint[] memory a) public {
        _original_Foo_foo(a);
        {
            emit AssertionFailed("HIT: 0: ");
            if (!(true)) {
                emit AssertionFailed("0: ");
                assert(false);
            }
        }
    }

    function _original_Foo_foo(uint[] memory a) private {}
}
```

Note that the event `emit AssertionFaield("HIT: 0: ")` is emitted regardless of whether the actual assertion failed, as long as its condition was evaluated.

We should add a command line flag `--cov-assertions` to emit these.

cd1m0:
This PR groups all the fixes/improvements done during engagements in October. Notably it:

1. Adds support for `#try` and `#required` - Closes #93 
2. Removes contract-level `#if_succeeds` annotations on view functions. Closes #99 
3. Fixes invalid code bug when interposing on constructors with super-constructor invocations. Closes #100 
4. Adds support in the type-checker for referring to library-defined constants. Closes #101 
5. Fixes invalid code when interposing on public state vars that override view functions. Closes #103 
6. Fixes incorrect code emitted when combining a `#forall` annotation with a `#require` annotation. Closes #104 
7. Adds a new `--cov-assertions` command line flag that emits a dummy assertion for each real assertion, that can be used to check if the real assertion was reached. Closes #106 
8. Bumps up solc-typed-ast to 7.0.2


src/instrumenter/instrument.ts
```diff
@@ -490,16 +505,59 @@ export function insertAnnotations(annotations: PropertyMetaData[], ctx: Transpil
...
+            annotation.type === AnnotationType.Try;
+
+            instrCtx.addAnnotationInstrumentation(annotation, stmt);
+            instrCtx.addAnnotationCheck(annotation, predicate);
+
+            return [stmt, !targetIsStmt];
+        } else {
+            const event = getAssertionFailedEvent(factory, contract);
+            return [emitAssert(ctx, predicate, annotation, event, emitStmt), false];
+        }
```
blitz-1306, line 556: Not a blocker, more a reminder for a following cleanups: consider to unnest as each `if`-`else` statement here contains `return` statement. So we may reduce code density in few lines here.

blitz-1306 (approved):
Wonderful job.

src/instrumenter/instrument.ts
cd1m0: Fixed in abe36c2.

cd1m0:
There are intermitten failures due to a segfault in `nyc` - our coverage computation too. Merging despite failures

blitz-1306 (approved):

Issue #123, opened 2021-11-18
Support `abi.encodePacked` in scribble


cd1m0:
Do as the title says. After fixing the following code should instrument:

```solidity
contract Foo {
   /// #if_succeeds abi.encodePacked(a).length > 0;
   function main(string memory a) public { }
}
```

Instead of throwing the below error:

```
tmp3.sol:2:37 TypeError: Builtin struct "abi" does not have a member "encodePacked"

In:

if_succeeds abi.encodePacked(a).length > 0;
```

cd1m0:
Actually we don't support `abi.encode` either in scribble. I think the reason we left those out originally was that they required support for variadic functions in the type-checker, and given that the only place where the were needed was in the `encode*`/`decode` family, we decided to skip them at the time. Welp, time to add that support I guess.

PR #124, opened 2021-11-19
Fixes for bugs #121 and #123

### LINKED ISSUE #121: Scribble grammar unnecessarily requires `forall` be wrapped in parenthesis
Given the following sample:

```solidity
contract Foo {
        /// #if_succeeds true && forall (uint i in arr) arr[i] > 1;
        function main(uint[] memory arr) public {}
}
```

scribble will produce the following syntax error:

```
tmp.sol:2:26 SyntaxError: Expected "!", "'", "(", "-", "0x", "\"", "hex", "int", "old", "u", [0-9], or whitespace but "f" found.

In:

/// #if_succeeds true && forall (uint i in arr) arr[i] > 1;
```

The issue is that in the grammar we specify precedence of operators with explicit rules, and the `For_All` rule is above the `LogicalANDExpression` rule. However I think its safe to move `For_All` at the bottom of the chain. Thus the bottom expression rule would become:

```
PrimaryExpression =
    HexLiteral
    / Identifier
    / Number
    / BooleanLiteral
    / StringLiteral
    / (
        "(" __ expr: Expression __ ")" { return expr; }
    )
    / (RESULT { return new SResult(location()); })
    / For_All
```

cd1m0:
This PR fixes #121  and #123 . Specifically:

1. Fix for #121  - tweak the grammar to support `forall`s inside boolean expressions without parenthesis. Add parsing tests
2. Fix for #123  - support the `abi.encode*` family of builtins. To do this adds a new internal `VariableTypes` type node, used to denote variable number of arguments for functions; Add typing definitions for the `abi.encode*` functions in `src/spec-lang/tc/builtins.ts`; Tweak `typecheck.ts` to handle type-checking of functions with variable number of arguments.

Note: We still don't support `abi.decode` as that additionally requires adding tuples to the scribble language.

### DIFF
--- a/src/spec-lang/tc/typecheck.ts
+++ b/src/spec-lang/tc/typecheck.ts
@@ -1409,7 +1409,10 @@ export function tcMemberAccess(expr: SMemberAccess, ctx: STypingCtx, typeEnv: Ty
         const baseLoc = baseT.location;
         const baseToT = baseT.to;
 
-        if (baseToT instanceof ArrayType && expr.member === "length") {
+        if (
+            (baseToT instanceof ArrayType || baseToT instanceof BytesType) &&
+            expr.member === "length"
+        ) {
             return new IntType(256, false);
         }
 
--- a/src/spec-lang/tc/typecheck.ts
+++ b/src/spec-lang/tc/typecheck.ts
@@ -1668,7 +1668,7 @@ function matchArguments(
         if (formalT instanceof VariableTypes) {
             assert(
                 i === funT.parameters.length - 1,
-                `Unexpected variable type not in last position of fun {1}`,
+                `Unexpected variable type not in last position of fun {0}`,
                 funT
             );
             return true;
--- a/src/spec-lang/tc/typecheck.ts
+++ b/src/spec-lang/tc/typecheck.ts
@@ -79,7 +79,7 @@ import {
     VarDefSite
 } from "../ast";
 import { BuiltinAddressMembers, BuiltinSymbols } from "./builtins";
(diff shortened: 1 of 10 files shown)


src/spec-lang/tc/typecheck.ts
```diff
@@ -1648,14 +1648,32 @@ function matchArguments(
...
         return false;
     }
 
     for (let i = 0; i < funT.parameters.length; i++) {
         const formalT = funT.parameters[i];
 
+        if (formalT instanceof VariableTypes) {
+            assert(
+                i === funT.parameters.length - 1,
+                `Unexpected variable type not in last position of fun {1}`,
```
blitz-1306, line 1671: Note that `assert()` placeholders [are zero-based](https://github.com/ConsenSys/solc-typed-ast/blob/ae66129bb5c8e49007ea56c882ff590df38a174d/src/misc/utils.ts#L26-L35). Message should rever `{0}` instead:
```
Unexpected variable type not in last position of fun {0}
```

blitz-1306 (commented):
Cool to see variadics support. Was always debating how to properly implement those. Good PR. One small note.

src/spec-lang/tc/typecheck.ts
cd1m0: Fixed in 8bb7a79.

blitz-1306 (approved):

blitz-1306 (approved):

blitz-1306 (approved):

cd1m0:
CI failure appears to be the intermittent segfault from nyc. Force-merging

Issue #126, opened 2021-12-02
`require` annotation keyword causing issues with openzeppelin


cd1m0:
When trying to instrument the below contract inheriting from open zeppelin:

```
// contracts/simple.sol
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.4;

import "@openzeppelin/contracts/token/ERC20/ERC20.sol";

/// #invariant unchecked_sum(_balances) == _totalSupply;
contract Simple is ERC20 {
    /**  @dev Token Constructor
     */     
    constructor() ERC20("Simple", "Simple") payable {
    }

```

Scribble throws the following error:

```
DEBUG:fuzzing-cli:Initializing configuration context
DEBUG:fuzzing-cli:Parsing config at .fuzz.yml
DEBUG:fuzzing-cli:Initializing tool name middleware with 0.8.1
Scribble has encountered an error (code: 1)
=====STDERR=====
@openzeppelin/contracts/token/ERC20/ERC20.sol:140:16 SyntaxError: Expected "!=", "%", "&", "&&", "*", "**", "+", "-", ".", "/", ";", "<", "<<", "<=", "==", "==>", ">", ">=", ">>", "?", "[", "^", "|", "||", or whitespace but "b" found.

In:


     * required by the EIP. See the note at the beginning of {ERC20}.
     *
     * Requirements:
     *
     * - `sender` and `recipient` cannot be the zero address.
     * - `sender` must have a balance of at least `amount`.
     * - the caller must have allowance for ``sender``'s tokens of at least
     * `amount`.
     */

=====STDOUT=====
```

The issue is that the regex pattern inside `findAnnotations` mistakenly considers the ` * required ....` to be the start of an annotation:

```
    const rx =
        /\s*(\*|\/\/\/)\s*#?(if_succeeds|if_updated|if_assigned|invariant|assert|try|require|define\s*[a-zA-Z0-9_]*\s*\([^)]*\)|macro\s*[a-zA-Z0-9_]*\s*\([^)]*\))/g;

    let match = rx.exec(meta.text);
    ...
```

Honestly I think the best solution is to finally disallow the (already deprecated) annotation keywords without a '#' (as tracked by #125).

If we decide to do this will close this as duplicate of that one.

blitz-1306:
There is also `try` that may be considered as regularly used keyword. So, this issue appears to be only a beginning.

cd1m0:
Fixed by #125 so closing as duplicate

Issue #125, opened 2021-11-23
Remove support for deprecated annotations without hashtag


cd1m0:
We have deprecated annotations that do not start with a hashatag since 2021-05-18 ( 03e036d8 ). We should now fully remove those to reduce code complexity.

cd1m0:
Fixed by #130

PR #130, opened 2021-12-10
[WIP] Remove support for deprecated annotations without a '#'

cd1m0:
This PR provides a fix for #125 and #126 . Specifically it:

1. Changes the scanning regex in `annotations.ts` and the grammar to ignore annotations without #
2. Simplifies the scanning regex in the cases of macro and define
3. Adds new code in `deprecated_warnings.ts` that looks for things that parse as annotations, but lack the # and prints warnings for those. Also does this for 2 more potential cases
4. Removes the deprecation notice printing code.

Note:
1. This depends on #97 and should be re-targeted to develop after #97 is merged.
2. After this is approved, I may do a bit of cleanup of the error throwing/printing code and group it as part of this PR. Will do a separate review for it.


src/instrumenter/deprecated_warnings.ts
```diff
@@ -0,0 +1,131 @@
...
+    units: SourceUnit[],
+    sources: SourceMap,
+    version: string
+): Warning[] {
+    const res: Warning[] = [];
+    const rxs: Array<[RegExp, string, (match: RegExpExecArray, txt: string) => string]> = [
+        // Old-style annotations without a '#'
+        [
+            /\s*(\*|\/\/\/)\s*[^#](if_succeeds|if_updated|if_assigned|invariant|assert|try|require|macro|define)/g,
+            "Deprecated annotation without '#' ignored. Please add the #.",
```
blitz-1306, line 76: WDYT if we make a suggestion:
> Detected possible annotation without #. If it is an annotation, then add `#` prefix. Otherwise it would be ignored.

blitz-1306 (approved):
Looks good. One suggestion (not a blocker).

src/instrumenter/deprecated_warnings.ts
cd1m0: Fixed in 99e3aec

cd1m0:
Force merging since:

1) This PR was already approved by @blitz-1306 , the approval was just negated by my re-base

2) Test run fine locally, the test failures in CI are due to our favourite little heisen-segfault in nyc. (We really gotta fix that...)

PR #133, opened 2021-12-13
Disarm without compiling

### LINKED ISSUE #119: Disarm fails when instrumented code fails to compile
If you have a contract that compiles before being instrumented but fails to compile after being instrumented, scribble will successfully arm the contract but fail to disarm it. This can leave a bit of a mess of armed files to clean up before you can fix or diagnose the issue.

### Example
Shadowed.sol
```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

interface IShadower {
    function shadowed(uint256 a) external;
    function shadowed(uint256 a, uint256 b) external;
}

/// #invariant {:msg "I0"} 0 == 0;
contract Scribbled {
    function entry() external {
        IShadower(address(0)).shadowed(1);
    }
}
```

```bash
solc Shadowed.sol
```
Succeeds.

```bash
scribble Shadowed.sol --output-mode files --arm
```
Succeeds, but leaves an instrumented `Shadowed.sol` that fails to compile.

In this case, it is because the function pointer passed to the `_callsite_` method cannot discern between the two versions of `IShadower.shadowed()`. As far as I know, this is a limitation of solidity rather than a scribble bug, but used here as an example.

```bash
scribble Shadowed.sol --output-mode files --disarm
```

Because the newly instrumented `Shadowed.sol` won't compile, this fails and leaves armed files that must be sorted out manually or reverted to a previous commit.

### Feature Request
- If it's not necessary to compile the armed contracts before disarming, that seems like an easy fix. Maybe this could be behind a `--disarm --force` flag.
- Attempt to compile the armed contracts and revert instrumenting if they fail to compile.

blitz-1306:
## Tasks
Fixes #119

## Changes
- [x] Added dependency `findup-sync`.
- [x] Introduced project root detection to store metadata file with "breadcrumbs". If there is no project root detected, then `--instrumentation-metadata-file` option is **required** to be specified to store metadata file.
- [x] Changed `scribble` executable file:
  - Check if some of the files are already instrumented before instumenting any of them.
  - Do not invoke compiler when disarming and rely on metadata file instead. Also, remove metadata file during disarming.
- [x] Fixed bug: macro file references in metadata file were marked with `.original`.
- [x] Fixed bug: some uninstrumented files during `--arm` (without `.original` equivalent)  in metadata file were marked as having an `.original` files.
- [x] Tweaked affected tests.

Regards.


src/bin/scribble.ts
```diff
@@ -550,8 +570,65 @@ if ("version" in options) {
...
+
+            originalFiles = new Set(original);
+            instrumentationFiles = new Set(instrumentation);
+        } else {
+            /**
+             * @todo: Can not throw here due to previous instrumentation will remain in projects.
+             */
+
+            originalFiles = new Set();
+            instrumentationFiles = new Set();
```
blitz-1306, line 613: This case is still unclear, especially if user update Scribble to latest while have some project armed and will not be able to disarm if do not have metadata file somewhere.

src/bin/scribble.ts
```diff
@@ -651,327 +703,319 @@ if ("version" in options) {
...
-        const factory = new ScribbleFactory(mergedCtx);
-
-        // Next we re-write the imports to fix broken alias references (Some
-        // Solidity versions have broken references imports).
-        mergedUnits.forEach((sourceUnit) => {
-            if (contentsMap.has(sourceUnit.absolutePath)) {
-                rewriteImports(sourceUnit, contentsMap, factory);
+        for (const unit of instrCtx.units) {
+            if (isInstrumented(unit.sourceEntryKey, instrumentationMarker)) {
+                error(`File "${unit.sourceEntryKey}" is already instrumented`);
```
blitz-1306, line 885: This seems more efficient then `if` branch above. @cd1m0 WDYT?

src/bin/scribble.ts
```diff
@@ -550,8 +570,65 @@ if ("version" in options) {
...
+
+        if (!options["keep-instrumented"]) {
+            for (const instrFileName of instrumentationFiles) {
+                remove(instrFileName, options);
+            }
+        }
+
+        if (isMetaDataFile) {
+            remove(metaDataFile, options);
+        }
```
blitz-1306, line 628: Note that disarming now also removes metadata file.

blitz-1306 (commented):
Some things to consider for reviewers.

src/bin/scribble.ts
```diff
@@ -550,8 +570,67 @@ if ("version" in options) {
...
     const ctxtsMap: Map<string, ASTContext> = new Map();
     const filesMap: Map<string, Map<string, string>> = new Map();
-    const originalFiles: Set<string> = new Set();
-    const instrumentationFiles: Set<string> = new Set();
+
+    let metaDataFile: string;
+    let originalFiles: Set<string>;
+    let instrumentationFiles: Set<string>;
+
+    if (options["instrumentation-metadata-file"]) {
```
cd1m0, line 578: Maybe move the logic for detecting the location of `instrumentation-metadata-file` in own helper function? Something like:

```
function detectInstrMetdataFilePath(options: any): string {
     ...
}```

File: src/bin/scribble.ts [hunk 5]
```diff
@@ -651,327 +705,314 @@ if ("version" in options) {
...
-            const metadata: any = generateInstrumentationMetadata(
-                instrCtx,
-                newSrcMap,
-                instrCtx.units,
-                modifiedFiles,
-                options["arm"] !== undefined,
-                pkg.version,
-                options["output"]
-            );
+    if (options["arm"] || options["instrumentation-metadata-file"]) {
```
cd1m0, line 1005: There is a small inconsistency here. In line 586 we throw an error if we can't detect a location for an instrumentation metadata file. We do this no matter the arguments to scribble. However we only really use the metadata files, if we have `--arm`, `--disarm` or `--instrumentation-metadata-file`. I am wondering if we should either:

a) Always emit the metadata file

 - or -
 
b) Only require an `--instrumentation-metadata-file` if we can't detect project root, and there is an `--arm` or `--disarm`

What do you think?

src/util/arming.ts
```diff
@@ -0,0 +1,55 @@
+import path from "path";
+import fse from "fs-extra";
+
+const findupSync = require("findup-sync");
+
+function getDirIfIsFile(p: string): string {
+    return fse.existsSync(p) && fse.statSync(p).isFile() ? path.dirname(p) : p;
+}
+
+export function getBasePath(paths: string[]): string {
```
cd1m0, line 10: This function will return "" if the user mixes absolute `/...` and relative `./...` paths. Perhaps it would be best if we normalize all paths to absolute in `detectProjectRoot` before we pass them to `getBasePath`?

cd1m0 (changes requested):
Awesome work! Just a couple of nits to fix

src/bin/scribble.ts
blitz-1306: Always emitting metadata does not match with the flat / JSON output. All operations are happening in memory and nothing is actually written to any files. Emitting the metadata file in such case causing confusion because Scribble starting to complaing that can not do same thing due to there is a metadata file, that can not be overwritten for different inputs. Currently Scribble follows option B from your list.

src/bin/scribble.ts
```diff
@@ -550,8 +570,67 @@ if ("version" in options) {
...
+        }
+
+        for (const originalFileName of originalFiles) {
+            if (originalFileName.endsWith(".sol.original")) {
+                move(originalFileName, originalFileName.replace(".sol.original", ".sol"), options);
+            }
+        }
+
+        if (!options["keep-instrumented"]) {
+            for (const instrFileName of instrumentationFiles) {
```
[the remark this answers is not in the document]
blitz-1306: `fs-extra` [does nothing](https://github.com/jprichardson/node-fs-extra/blob/master/docs/remove-sync.md) in case if there is no file to remove.

conejo1130 (changes requested):
#102

src/util/arming.ts
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/133/commits/62dfc4eda8b6d526ae493944f6749a62b738b079

src/bin/scribble.ts
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/133/commits/62dfc4eda8b6d526ae493944f6749a62b738b079

src/bin/scribble.ts
```diff
@@ -550,8 +570,67 @@ if ("version" in options) {
...
+             * @todo: Can not throw here due to previous instrumentation will remain in projects.
+             */
+
+            originalFiles = new Set();
+            instrumentationFiles = new Set();
+        }
+
+        for (const originalFileName of originalFiles) {
+            if (originalFileName.endsWith(".sol.original")) {
+                move(originalFileName, originalFileName.replace(".sol.original", ".sol"), options);
```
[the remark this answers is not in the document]
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/133/commits/62dfc4eda8b6d526ae493944f6749a62b738b079

src/bin/scribble.ts
```diff
@@ -461,9 +462,27 @@ function remove(filePath: string, options: any): void {
...
+        return false;
+    }
+
+    const source = fse.readFileSync(filePath, { encoding: "utf-8" });
+
+    return source.startsWith(prefix);
+}
+
+function getFilesInMetaData(fileName: string): { original: string[]; instrumentation: string[] } {
+    const metaData = fse.readJsonSync(fileName, { encoding: "utf-8" });
```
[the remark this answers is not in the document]
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/133/commits/62dfc4eda8b6d526ae493944f6749a62b738b079. A marked as TODO to pick a package for proper [JSON schema validation](https://json-schema.org/implementations.html#validator-javascript).

src/bin/scribble.ts
```diff
@@ -550,8 +570,67 @@ if ("version" in options) {
...
+    if (options["instrumentation-metadata-file"]) {
+        metaDataFile = options["instrumentation-metadata-file"];
+    } else {
+        const projectRoot = detectProjectRoot(targets);
+
+        if (projectRoot !== undefined) {
+            metaDataFile = join(projectRoot, "instrumentation.scribble.json");
+        } else {
+            error(
+                'Unable to detect project root to place instrumentation metadata file. Use "--instrumentation-metadata-file" option to explictly specify location.'
```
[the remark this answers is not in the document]
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/133/commits/62dfc4eda8b6d526ae493944f6749a62b738b079

src/bin/scribble.ts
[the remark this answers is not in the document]
blitz-1306: We should consider using https://github.com/ajv-validator/ajv for this in the future iterations.

src/bin/scribble.ts
cd1m0: Thats fair. I agree with you.

cd1m0 (approved):
LGTM! Just needs a rebase

src/bin/scribble.ts
cd1m0: lgtn

PR #134, opened 2021-12-21
Fix for #132

### LINKED ISSUE #132: Multitple 'AssertionFailed' events inserted in inheritance
Given the following samples:

base.sol:
```
pragma solidity 0.8.7;

/// #invariant true;
contract Base {
    function bar() public {}
}
```

child.sol:
```
pragma solidity 0.8.7;

import "./base.sol";

/// #invariant true;
contract Child is Base {
    function foo() public {}
}
```

scribble produces code that doesn't compile:

```
$ scribble child.sol  base.sol
/// #invariant true;
contract Base is __scribble_ReentrancyUtils {
...
}

/// #invariant true;
contract Child is __scribble_ReentrancyUtils, Base {
...
}
```

The issue is that both the parent class (Base) and subclass (Child) define the same event `AssertionFailed`.
This error doesn't occur if we switch the order of the arguments, or if we put them in the two contracts in the same file.

The problem stems from the fact that the logic that determines where `AssertionFailed` are instered depends on the order of visiting contracts:

```jsx
export function insertAnnotations(annotations: PropertyMetaData[], ctx: TranspilingContext): void {
...
        const event = getAssertionFailedEvent(factory, contract);
        return [emitAssert(ctx, predicate, annotation, event, emitStmt), false];
...
}

```jsx
export function getAssertionFailedEvent(
    factory: ASTNodeFactory,
    contract: ContractDefinition
): EventDefinition {
    const events = resolveByName(contract, EventDefinition, "AssertionFailed");

    if (events.length > 0) {
        return events[0];
    }
    ...
```

Basically in which-ever contract we first add annotations, we will also emit the event. This is a problem when there are multiple contracts in a single inheritance chain that have the require an event, and when we first happen to visit a child contract.

A cleaner solution would be to always insert `AssertionFailed` and `AssertionFailedData` events in the `__scribble_ReentrancyUtils` contract and make all contracts inherit from it.

cd1m0:
This pr fixes #132 

Specifically it:
- moves the `AssertionFailed` and `AssertionFailedData` event definitions inside the `__scribble_ReentrancyUtils` contract.
- changes the instrumentation logic such that any contract using those events gets `__scribble_ReentrancyUtils` as a base contract (in the past it was just the contracts with `#invariant` annotations)


src/instrumenter/instrumentation_context.ts
```diff
@@ -488,6 +488,67 @@ export class InstrumentationContext {
...
+            ctx instanceof ContractDefinition
+                ? ctx
+                : ctx.getClosestParentByType(ContractDefinition);
+
+        assert(containingContract !== undefined, `Node {0} not under a contract`, ctx);
+        this.addScribbleUtils(containingContract);
+
+        return single(
+            this.utilsContract.vEvents.filter((evt) => evt.name === "AssertionFailedData")
+        );
```
blitz-1306, line 549: These two methods only differs by the name of event: `AssertionFailedData` or `AssertionFailed`. Maybe introduce helper to reduce code duplication here?

test/samples/assert_forall.instrumented.sol
```diff
@@ -2,9 +2,16 @@
 /// Use --disarm prior to make any changes.
 pragma solidity 0.8.7;
 
-contract AssertForall {
+/// Utility contract holding a stack counter
+contract __scribble_ReentrancyUtils {
     event AssertionFailed(string message);
 
+    event AssertionFailedData(int eventId, bytes encodingData);
```
blitz-1306, line 9: This event is not used in the instrumented contract. We no longer generate this event on-demand, right. Is it a current expectation?

blitz-1306 (approved):
Good work. One question and one suggestion.

test/samples/assert_forall.instrumented.sol
cd1m0: Yep this is intended. Simpler logic to always emit it

src/instrumenter/instrumentation_context.ts
cd1m0: Fixed in 8de918c

blitz-1306 (approved):
Looks good.

PR #137, opened 2022-01-18
Support 0.8.8 0.8.9 0.8.10

cd1m0:
This PR adds support for Solidity versions 0.8.8 - 0.8.10. Specifically it adds support for the following language features:

- User defined value types
- type(...).{min, max} for enum types, as well as int types (we didnt support these before)
- type(...).{name, runtimeCode, creationCode, interfaceId} for contract and interface types.


src/spec-lang/tc/builtins.ts
```diff
@@ -212,3 +213,30 @@ export const BuiltinAddressMembers = new Map<string, TypeNode | [TypeNode, strin
...
+        ["creationCode", new PointerType(new BytesType(), DataLocation.Memory)],
+        ["runtimeCode", new PointerType(new BytesType(), DataLocation.Memory)]
+    ])
+);
+
+export const InterfaceTypeMembers = new BuiltinStructType(
+    "<type(interface)>",
+    new Map<string, TypeNode>([
+        ["name", new PointerType(new StringType(), DataLocation.Memory)],
+        ["interfaceId", new FixedBytesType(4)]
```
blitz-1306, line 230: Available since [Solidity v0.6.7](https://github.com/ethereum/solidity/releases/tag/v0.6.7). I guess it is worth to specify minimal required version.

src/spec-lang/tc/builtins.ts
```diff
@@ -212,3 +213,30 @@ export const BuiltinAddressMembers = new Map<string, TypeNode | [TypeNode, strin
...
+    ])
+);
+
+export const NumberLikeTypeMembers = (type: IntType | UserDefinedType): BuiltinStructType => {
+    return new BuiltinStructType(
+        "<type(number-like)>",
+        new Map<string, TypeNode>([
+            ["min", type],
+            ["max", type]
```
blitz-1306, line 239: Available since [Solidity v0.6.8](https://github.com/ethereum/solidity/releases/tag/v0.6.8). I guess it is worth to specify minimal required version.

src/spec-lang/tc/typecheck.ts
```diff
@@ -1470,6 +1475,34 @@ export function tcMemberAccess(expr: SMemberAccess, ctx: STypingCtx, typeEnv: Ty
...
+
+        if (expr.member === "wrap") {
+            return new FunctionType(
+                "wrap",
+                [typeNameToTypeNode(underlyingType)],
+                [baseT.type],
+                FunctionVisibility.Default,
+                FunctionStateMutability.Pure
+            );
+        } else if (expr.member === "unwrap") {
```
blitz-1306, line 1495: `else if` could be an `if` fron new line - would be a bit easier to read/scan the code flow

blitz-1306 (commented):
Good job. Few questions and suggestions.

src/rewriter/merge.ts
```diff
@@ -283,8 +284,10 @@ export function merge(groups: SourceUnit[][]): [SourceUnit[], ASTContext] {
                 child.scope = getNew(child.scope).id;
             } else if (child instanceof ErrorDefinition) {
                 // Nothing to do
-            } else if (child instanceof Block) {
+            } else if (child instanceof UserDefinedValueTypeDefinition) {
+                // Nothing to do
```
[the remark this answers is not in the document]
cd1m0: I would like to leave those in, as a form of documentation. I.e. having 

```
            } else if (child instanceof ErrorDefinition) {
                // Nothing to do
```

Is a way of marking that I have thought about this AST node, and think that nothing has to be tweaked on it when merging.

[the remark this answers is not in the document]
cd1m0: Also this way when a new type of node gets added, we get a loud failure forcing us to think through if it need to be handled specially during merging.

src/spec-lang/tc/builtins.ts
```diff
@@ -212,3 +213,30 @@ export const BuiltinAddressMembers = new Map<string, TypeNode | [TypeNode, strin
     ],
     ["code", [new PointerType(new BytesType(), DataLocation.Memory), ">=0.8.0"]]
 ]);
+
+export const ContractTypeMembers = new BuiltinStructType(
```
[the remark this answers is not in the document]
cd1m0: Fixed in 9555f70.

src/spec-lang/tc/builtins.ts
cd1m0: Good catch. Fixed in Fixed in 9555f70.

src/spec-lang/tc/builtins.ts
cd1m0: Good catch. Fixed in Fixed in 9555f70.

src/spec-lang/tc/typecheck.ts
cd1m0: Fixed in 9555f70.

blitz-1306 (approved):
LGTM

PR #138, opened 2022-01-20
Support annotation label short-hand with just a string

cd1m0:
This adds a small shorthand for annotation labels. Instead of having to write this:

```
#invariant {:msg "Invariant1"} ...;
```

Now you can just write:

```
#invariant "Invariant1" ...;
```

The old syntax is still supported.


blitz-1306 (approved):
LGTM

PR #139, opened 2022-01-26
Add standard library support

cd1m0:
This PR adds a built-in standard library of macros that is shipped with the normal scribble release. You can refer to those macros without having to specify an explicit `--macro-path` command line option. Initially the library contains 3 macros:

- ERC20 macro. You can invoke it for any standard delfationary (fixed-supply) ERC20 contract with the `#macro erc20(balances, allowances, totalSupply);` contract-level annotation. Note that the `balances`, `allowances` and `totalSupply` are names of expected state variables in the annotated contract

- ERC721 macro. You can invoke it for any standard ERC721 contract with the `#macro erc721();` annotation

- Ownable macro. You can invoke it for the standard Ownable contract with the `#macro ownable(owner);` annotation where `owner` is the name of the state variable storing the owner.


src/instrumenter/annotations.ts
```diff
@@ -502,20 +502,48 @@ function processMacroAnnotations(
...
-                        meta.parsedAnnot.src as Range,
-                        meta.target
-                    );
-                }
+                let target: AnnotationTarget;
 
-                const target = targets[0];
+                if (name === "<contract>") {
+                    target = scope;
+                    assert(target instanceof ContractDefinition, ``);
```
blitz-1306, line 509: Empty error message in `assert()`.

src/instrumenter/annotations.ts
```diff
@@ -502,20 +502,48 @@ function processMacroAnnotations(
...
-                        } for ${meta.original}`,
-                        meta.original,
-                        meta.parsedAnnot.src as Range,
-                        meta.target
-                    );
-                }
+                let target: AnnotationTarget;
 
-                const target = targets[0];
+                if (name === "<contract>") {
```
blitz-1306, line 507: I think this entire `if` block can be separated to a helper function with earlier returns. Otherwise `processMacroAnnotations()` becaming complex with e lot of nesting. It would be more readable if we keep it less complex.

test/integration/src2srcmap.spec.ts
```diff
@@ -373,7 +374,7 @@ describe("Src2src map test", () => {
 
                 // Need to skip some compiler-generated bytecode. Specifically any
                 // bytecode related to entire functions, contracts or (as of 0.8.6) the entire body
-                // of a function
+                // of a function. Additionally we skip anything iside of an inline assembly node
```
blitz-1306, line 377: Typo in word `inside`

test/integration/utils.ts
```diff
@@ -135,13 +135,7 @@ export function scribble(fileName: string | string[], ...args: string[]): string
...
 }
 
-const defaultArgsMap = new Map<string, string[]>([
-    [
-        "test/samples/ownable_macro.sol",
-        ["--macro-path", "test/samples/macros/ownable.scribble.yaml"]
-    ],
-    ["test/samples/erc20_macro.sol", ["--macro-path", "test/samples/macros"]]
-]);
+const defaultArgsMap = new Map<string, string[]>([]);
```
blitz-1306, line 138: I do not think there is a need for this variable, as it used for reading only and contains no values.

blitz-1306 (commented):
Good job. A few ideas and suggestions.

src/instrumenter/annotations.ts
cd1m0: Fixed in 2e5e76b.

src/instrumenter/annotations.ts
cd1m0: Fixed in 2e5e76b.

test/integration/src2srcmap.spec.ts
cd1m0: Fixed in 2e5e76b.

test/integration/utils.ts
cd1m0: Fixed in 2e5e76b.

blitz-1306 (approved):
Looks good to me

PR #143, opened 2022-02-03
Use native compilers for AST building

blitz-1306:
## Preface
This PR allows to use native Solidity compilers to build and process AST.

## Changes
- [x] Updated dependencies to latest.
- [x] Use native compilers instead of WASM for AST building and processing.
- [x] Introduced compiler kind configuring via CLI option `--compiler-kind` and environment variable `SCRIBBLE_DEFAULT_COMPILER_KIND`. Precedence is following:
  - `--compiler-kind` is primary;
  - `SCRIBBLE_DEFAULT_COMPILER_KIND` is secondary;
  - If `getCompilerPrefixForOs()` result is one of `linux-amd64`, `windows-amd64` or `macosx-amd64`, then `CompilerKind.Native`. Otherwise the fallback value is `CompilerKind.WASM`.
- [x] Tweaked executable and tests to be compatible with `async` execution.

## Notes
- It seems that **solc-typed-ast** v8.0.0 adds `./` prefix to all relative input paths, including entry file path. Such behavior affected a lot of artifacts and snapshots, so it probably would also affect dependent components. We better to assess possible implications of this change and consider if it worth fixing in the **solc-typed-ast** itself.
  - This is addressed in https://github.com/ConsenSys/solc-typed-ast/pull/93.

Regards.


src/bin/scribble.ts
```diff
@@ -540,511 +581,531 @@ function pickVersion(versionUsedMap: Map<string, string>): string {
...
-             * the inability to merge the id-spaces of the nodes of different compilation results.
-             */
-            compilerVersionUsedMap.set(target, compilerVersionUsed);
-            groupsMap.set(target, originalUnits);
-            ctxtsMap.set(target, ctx);
-            filesMap.set(target, targetResult.files);
-        } catch (e) {
-            console.error(e);
+                const originalUnits = reader.read(targetResult.data, undefined, targetResult.files);
+                /**
```
cd1m0, line 769: Nit: This comment is stale. Should delete

src/bin/scribble.ts
```diff
@@ -540,511 +581,531 @@ function pickVersion(versionUsedMap: Map<string, string>): string {
...
+                process.exit(1);
+            }
         }
-    }
 
-    /**
-     * Without --disarm we need to instrument and output something.
-     */
-    const contentsMap: SourceMap = new Map();
+        /**
```
cd1m0, line 784: Nit: This comment is stale. Should delete

test/unit/custom_maps.spec.ts
```diff
@@ -600,7 +606,7 @@ describe("Interposing on a map", () => {
                 ]
             };
 
-            executeTestSuiteInternal("foo.json", cfg, version);
+            executeTestSuiteInternal(cfg, version);
```
cd1m0: Does this need an await?

cd1m0 (changes requested):
Just 1 question and 2 nits. Will fix then its shortly

src/bin/scribble.ts
cd1m0: Fixed in [839523e](https://github.com/ConsenSys/scribble/pull/143/commits/839523e6d9a55d3032a411ed5152ef90b9d21a35)

src/bin/scribble.ts
cd1m0: Fixed in [839523e](https://github.com/ConsenSys/scribble/pull/143/commits/839523e6d9a55d3032a411ed5152ef90b9d21a35)

test/unit/custom_maps.spec.ts
cd1m0: Added await in [d19be01](https://github.com/ConsenSys/scribble/pull/143/commits/d19be011aa01087d036d8c863e32aa787b0afff4)

cd1m0 (dismissed):
lgtm

cd1m0 (changes requested):
After discussion with @blitz-1306 we decided to make the logic for selecting default compiler kind a little more robust. Specifically the selection logic should be:

   1. If there is `--compiler-kind` specified use that
   2. Otherwise if there is an env variable `SCRIBBLE_DEFAULT_COMPILER_KIND`, then use that
   3. Finally if no options are given use `CompilerKind.Native` IF the return value of `getCompilerPrefixForOs()` is one of the 3 supported platforms - `"linux-amd64"`, `"windows-amd64"` and `"macosx-amd64"`. Otherwise we use `CompilerKind.WASM`.

cd1m0 (approved):
LGTM

PR #157, opened 2022-03-03
[EXPERIMENTAL] Support local ghost variables with `let` annotations

cd1m0:
**[WARNING]: This feature is experimental, and may be changed and potentially removed in the future. Feel free to explore it, but please don't depend on it for core annotations until its stabilized.**

This PR provides a proof-of-concept implementation of local ghost variables using let-bindings.

The motivation is that sometimes scribble is not expressive enough to express certain properties at the function boundary, and instead we might want to embed annotations in the middle of the function. We already support the `#assert` annotation that can be inserted before any statement, however that is not enough to talk about the change of variables before and after some statement. For example consider the following loop:

```
function batchTransfer(address[] memory receivers, uint[] memory amounts) {
    for (uint i = 0; i < receivers.length; i++) {
        address receiver = receivers[i];
        uint amount = amounts[i];
        token.transfer(receiver, amount);
    }
...
```

We might want to check the spec that "after all transfers, the balances of recipients increase correctly. Naively we might try to express it with the following if_succeeds annotation:

```
/// #if_succeeds forall(uint i in receivers) token.balanceOf(receivers[i]) == old(token.balanceOf(receivers[i])) + amounts[i];
function batchTransfer(address[] memory receivers, uint[] memory amounts) {
```

However this will not be correct if some address is repeated in `receivers`! So instead we might want to try and check that each individual transfer succeeds by adding an `#assert` in the body of the loop after the `token.transfer`. However `#assert` cannot talk about the **old** value of the balance. This is where #let comes in:

```
    for (uint i = 0; i < receivers.length; i++) {
        address receiver = receivers[i];
        uint amount = amounts[i];
        /// #let oldBalance := token.balanceOf(receiver);
        token.transfer(receiver, amount);
        /// #assert token.balanceOf(receiver) == oldBalance + amount;
        0;
    }
```

In the above specification, the `#let` is used to add a ghost local variable `oldBalance`, that can be accessed by the following `#assert`. There are serveral things to note about ghost local variables:

1. Ghost local variables follow the same scoping rules as solidity variables.
2. Ghost local variables are immutable - you can't reassign them. (and thus you can't use them for example to accumulate some sum)
3. Ghost local variables must have unique names that don't alias ANY other identifier in the current program (to avoid confusion)


blitz-1306 (dismissed):
Looks good.

wuestholz (dismissed):
LGTM! 👍

PR #162, opened 2022-03-30
Better support for Solidity 0.8.13

blitz-1306:
## Preface
This PR is addressing review consideration from [here](https://github.com/ConsenSys/scribble/pull/161#pullrequestreview-923976302).

## Dependencies
- This PR depends on #161.

## Changes
- [x] Updated dependencies to latest and fixed `npm audit` issue.
- [x] Fix `merge()` when free function is imported after Solidity 0.8.13 (it was chashing on the added test sample).
- [x] Fix flattening to support renaming member accesses, that are refer to renamed free functions (bound via using-for directive).
- [x] Fix `getAssignments()` to also respect bound free functions.
- [x] Added mutlifile sample to track the use-case.

Regards.

PR #164, opened 2022-03-31
Utilize multifile compiling

blitz-1306:
## Dependencies
- This PR depends on #162 (and should be reviewed/merged after it).

## Changes
- [x] Tweaked Scribble executable to utilize multifile compiling from **solc-typed-ast**.
- [x] Removed source unit merging logic - this is done by compiler on its own.
- [x] Updated affected test samples.

## Notes
- **Multifile compiling is supported only for `source` input mode.** The `json` input mode assumes that all necessary sources are already in the input file. Merging multiple JSON files does not seem to be a great idea due to possible key collistions and amount of possibe input scenarios. So there is still requirement to have a single input file when input mode is set to `json`.
- I suggest to pay additional attention to changed flattened samples. Some node ids was changed. Seems to be "okay" but I'm not completely sure that it really is "okay".

Regards.


src/bin/scribble.ts
```diff
@@ -151,8 +150,17 @@ function ppWarning(warn: Warning): string[] {
     return [`${ppLoc(start)} Warning: ${warn.msg}`, getSrcLine(warn.location)];
 }
 
+function assertIsFile(fileName: string): void {
+    assert(fileName !== "", "Path is empty");
+    assert(fse.existsSync(fileName), "Path not found: " + fileName);
+
+    const stats = fse.statSync(fileName);
```
cd1m0, line 157: This is not a blocker - just a small optimization. We can remove the call to `fse.existsSync` here by doing:
```
assert(fileName !== "", "Path is empty");
try {
        const stats = fse.statSync(fileName);
        assert(stats.isFile(), "Target is not a file: " + fileName);
} catch (e) {
         assert(false, e.code === 'ENOENT' ? ("Path not found: " + fileName) :  e.message);
}
```

src/bin/scribble.ts
```diff
@@ -891,44 +887,21 @@ function pickVersion(versionUsedMap: Map<string, string>): string {
...
-                isSane(mergedUnit, mergedCtx),
-                `Merged unit ${mergedUnit.absolutePath} is insane`
-            );
-        }
-
-        for (const target of targets) {
-            const units = groupsMap.get(target) as SourceUnit[];
-            const files = filesMap.get(target) as Map<string, string>;
+        for (const unit of units) {
+            assert(isSane(unit, astCtx), `Merged unit ${unit.absolutePath} is insane`);
```
cd1m0, line 892: I don't think we need this assert anymore

cd1m0 (changes requested):
Good job! Love commits that delete code. Two small nits - one is not a blocker

src/bin/scribble.ts
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/164/commits/42b350e912bd4eef2f59dc60d3bdfcb6f27e67d3

src/bin/scribble.ts
blitz-1306: Addressed in https://github.com/ConsenSys/scribble/pull/164/commits/42b350e912bd4eef2f59dc60d3bdfcb6f27e67d3

cd1m0 (approved):
LGTM

PR #165, opened 2022-03-31
Fix for #163

### LINKED ISSUE #163: Scribble doesn't correctly handle private visiblity of state vars
Scribble will accept and instrument the following code, even though it shouldn't:

```solidity
contract Base {
    uint private x;
}

contract Child is Base {
     /// if_succeeds x == 42;
     function foo() public {}
}
```

The problem is that the annotation in the `Child` contract refers to a **private** state variable of the parent contract. Those cannot be accessed in the child class. This occurs for example when you have a token contract inheriting from @openzeppelin's ERC20 contract, that talks about the _balances state var.

There are 2 possible solutions:

1. Fix the type-checker to reject annotations where we are talking about private state variables of base contracts
2. Change the instrumentation to silently change the visiblity of referenced private state variables into 'internal', thus allowing annotations to talk about private state of base contracts.

cd1m0:
This PR includes:

* bump to solc-typed-ast version 9.1.2 which fixes #163  (for now - we may revisit this fix in the future under a different issue)
* small followup fix to the previous PR that fixed  https://github.com/ConsenSys/scribble/pull/161 - apparently when the remapping involved non-normalized elements such as './' or '//' that could confuse our logic for selecting where to place the utils unit

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
                 "glob": "^7.2.0",
                 "logplease": "^1.2.15",
                 "semver": "^7.3.5",
-                "solc-typed-ast": "^9.1.1",
+                "solc-typed-ast": "^9.1.2",
                 "src-location": "^1.1.0",
                 "yaml": "^1.10.2"
             },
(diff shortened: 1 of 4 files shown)


blitz-1306 (approved):
Looks good.

PR #173, opened 2022-05-24
Fix fixed size bytes width literal type

### LINKED ISSUE #159: Scribbles grammar generates invalid FixedBytes type objects
Given the following code:

```solidity
pragma solidity 0.8.12;

contract Foo {
    /// #if_succeeds forall(bytes32 x in hashes) x != bytes32(0x0);
    function foo(bytes32[] memory hashes) public {}
}
```

Scribble outputs the following error:

```
tmp.sol:4:36 TypeError: The type bytes[object Object] of the iterator variable x is not compatible with iterator type uint of hashes.
tmp.sol:4:36:
    /// #if_succeeds forall(bytes32 x in hashes) x != bytes32(0x0);
```

Note the `[object Object]` part of the string. This comes from the `FixedBytesType` pretty printer, and indicates that the number field of the `FixedBytesType` is invalid. The issue is on line 702 of the [scribble grammar](https://github.com/ConsenSys/scribble/blob/develop/src/spec-lang/expr_grammar.pegjs#L702):

```pegjs
FixedSizeBytesType =
    BYTES width: Number {
        return new FixedBytesType(width, makeRange(location(), options as ParseOptions));
    }
    / BYTE {
        return new FixedBytesType(1, makeRange(location(), options as ParseOptions));
    }
```

Specifically the `width: Number` grammar member returns an `SNumber` object, while the `FixedBytesType` constructor expects a simple JS number.

blitz-1306:
## Tasks
This PR fixes #159 

## Changes
- [x] Fixed audit issue in dependencies.
- [x] Fixed grammar issue with fixed size bytes.

Regards.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
     "packages": {
         "": {
             "name": "eth-scribble",
-            "version": "0.6.7",
+            "version": "0.6.8",
             "license": "SEE LICENSE IN <LICENSE>",
             "dependencies": {
                 "big-integer": "^1.6.51",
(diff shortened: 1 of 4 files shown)


cd1m0 (approved):
Good catch!

PR #177, opened 2022-06-07
Allow structs to use bound library methods and fix imports in executable

### LINKED ISSUE #174: using Counters.Counter
Hi,

When using OpenZeppelin's counters, scribble is unable to understand that for example, `current()` is a function and says Counters.Counter doesn't have property `current`.

```
using Counters for Counters.Counter;
Counters.Counter private _tokenIdCounter;
```

For example:

`///#if_succeeds old(tokenCounter.current()) + 1 == tokenCounter.current();`

Can we look into this?

Thank you

### LINKED ISSUE #175: Cannot find module 'path/posix'
After installing scribble (version 0.6.9) and trying to run it, I got the following error: `Error: Cannot find module 'path/posix'`. However, I do have `path/posix` installed as a global package in my nodejs setup.

After inspecting `eth-scribble/dist/bin/scribble.js`, I noticed that the require statement for that module is incorrect. It currently states:

```
const posix_1 = require("path/posix");
```

whereas it should be

```
const posix_1 = require("path-posix");
```

After manually patching the require statement as indicated above, scribble worked as expected.

blitz-1306:
## Tasks
- Fixes #174
- Fixes #175 

## Changes
- [x] Allowed struct expressions in annotations to use bound library methods (weakened the check, that was forcing to use only direct fields).
- [x] Fixed imports in Scribble executable.

Regards.

### DIFF
--- a/src/bin/scribble.ts
+++ b/src/bin/scribble.ts
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
 import fse from "fs-extra";
-import { dirname, join, relative, resolve } from "path";
-import { normalize } from "path/posix";
+import { dirname, join, relative, resolve, normalize } from "path";
 import {
     assert,
     ASTContext,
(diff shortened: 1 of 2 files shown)


cd1m0 (approved):
LGTM!

PR #178, opened 2022-06-17
Avoid invariants double check in constructors

### LINKED ISSUE #140: Double-checking of state invariants in constructor
If a contract has both invariants as well as function annotations on the constructor, we will check the contract invariants twice at the end of the constructor. This is unnnecessary. For example given the below sample:

```
/// #invariant true;
contract Foo {

 /// #if_succeeds false;
 constructor(uint x) public {
 }
}
```

Running `scribble sample.sol` will produce something like:

```
...
    constructor(uint x) {
        vars1 memory _v;
        _v.__scribble_check_invs_at_end = __scribble_out_of_contract;
        _original_Foo_constructor(x);
        unchecked {
            if (!(false)) {
                emit AssertionFailed("1: ");
                assert(false);
            }
        }
        if (_v.__scribble_check_invs_at_end) __scribble_check_state_invariants();
        __scribble_out_of_contract = _v.__scribble_check_invs_at_end;
    }

    function _original_Foo_constructor(uint x) private {
        __scribble_check_state_invariants();
        __scribble_out_of_contract = true;
    }
 ....
```

Note that `__scribble_check_state_invariants` is called both from `_original_Foo_constructor` as well as `constructor`. This only happens for constructors.

blitz-1306:
## Tasks
- Fixes #140

## Changes
- [x] Skip instrumenting constructors by default logic if we have attached annotations. This is to avoid adding state invariants double-check, as we add checkers in two places for this case.
- [x] Added related test sample. 

Regards.


cd1m0 (approved):
LGTM

PR #180, opened 2022-06-27
Support user-defined constants

### LINKED ISSUE #116: Add support for scribble constants in the language
In properties we've written there is often some constant  that keeps on repeating. (e.g. number of decimals in a Ray/Wad, number of seconds in a day, etc). It would make code cleaner if we could define Scribble constants similarly to how we can define user helper functions in scribble. I.e. something like:

```solidity
/// #define SECONDS_PER_DAY: uint  = 60*60*24;
contract Foo {
    uint timestamp;
    
    /// #limit {:msg "no more than a day passed"} block.timestamp - timestamp < SECONDS_PER_DAY;
    function bar() public {}
}
```

There is still the issue here that constants would be defined per contract, while sometimes we may want to use them from another contract. No clear story for that yet, but the incremental improvement might be worth it.

blitz-1306:
## Tasks
- Resolves #116

## Status
This PR is still unfinished (WIP). See "notes" section. Also implementation is a bit _naive_.

## Changes
- [x] Introduced constant annotations with `const` keyword. First I thought that it would be better to use `define`, however it would cause much alterations, including annotation filtering and distincting functions from consts. In the end I figured that we can use other keyword instead.
- [x] Constants are inserted to Scribble **utils** contract and are scoped by `SourceUnit`.
- [x] Simplified `TypeEnv` and dedicated separate classes `UserConstantsScope` and `UserFunctionsScope` to handle user-defined funcs and consts.
- [x] Added simple test samples.

## Notes
- Consts are order-dependent for now. E.g. developer can not use `Y := X + 1` and then `X := 1`. For that we would either require a [dependency graph](https://deepsource.io/glossary/dependency-graph/) or [hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting). I see both of them as a complication (as developers may be consfused by not paying much attention to maintaining proper order of constants), so did not invested much time into this.
- Consts can not yet be imported to other `SourceUnit`s. This is to simplify MVP feature implementation. We can introduce imports later.
- Consts are allowing some constructs, that should not be available. I didn't yet found a way to restrict usable expression to allow only literals and other consts.
- I guess we need more tests, considering shadowing and other possible edge-cases.
- There are several TODOs, that worth discussing first (that may happen during PR review).

Regards.


test/multifile_samples/node_modules_erc20/contracts/Foo.sol.instrumented
```diff
@@ -0,0 +1,30 @@
+/// This file is auto-generated by Scribble and shouldn't be edited directly.
```
cd1m0, line 1: Why is this file part of the PR? Did I forget it originally, or does the PR just need to be rebased?

test/samples/consts_simple.instrumented.sol
```diff
@@ -0,0 +1,49 @@
...
+                emit AssertionFailed("2: ");
+                assert(false);
+            }
+        }
+    }
+
+    function _original_Some_testHD() private {}
+}
+
+/// #const uint256 W := D * 7;
```
cd1m0, line 36: Hmmm, do we have a guarantee that contracts in a single file will be processed by scribble in the same order as they are defined in the file? I.e. is it possible for scribble to process `Other` before processing `Some`?

cd1m0 (changes requested):
Awesome work! A couple of small things to reconsider. Also you are correct that we need to tweak semantic checking to disallow some constructs for const definitions (e.g. old()). Also, I need to do another pass over the typechecking code, as that part is a bit mroe complex.

src/bin/scribble.ts
```diff
@@ -992,9 +997,16 @@ function loadInstrMetaData(fileName: string): InstrumentationMetaData {
...
+            /**
+             * Type check
+             */
             tcUnits(units, annotMap, typeEnv);
-            // Semantic check
+
+            /**
+             * Semantic check
+             *
+             * @todo #const should have more restrictions, not allowing certain expressions to be used.
```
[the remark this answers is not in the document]
blitz-1306: For an instance, using certain builtins, like `msg.sender` or costructs, maybe... Unsure about whole list. Will try to make descriptive sample.

src/instrumenter/instrument.ts
```diff
@@ -670,6 +679,69 @@ function makeUserFunctions(
...
+        const constType = constDef.formalType;
+
+        const userConst = factory.makeVariableDeclaration(
+            true,
+            false,
+            nameGen.getFresh(`${constDef.name.name}_${constUnit.id}_`),
+            ctx.utilsContract.id,
+            true,
+            getTypeLocation(constType),
+            StateVariableVisibility.Public,
```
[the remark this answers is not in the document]
blitz-1306: Fixed in https://github.com/ConsenSys/scribble/pull/180/commits/c5e0537b6f1900ecd339ed2173c5000b362e6bd8

src/instrumenter/transpile.ts
```diff
@@ -326,9 +328,27 @@ function transpileId(expr: SId, ctx: TranspilingContext): Expression {
...
+        const transpiledDecl = ctx.instrCtx.userConstants.get(expr.defSite);
+
+        assert(
+            transpiledDecl !== undefined,
+            "Missing transpiled version of user constant {0}",
+            expr.defSite
+        );
+
+        /**
+         * @todo Do we need to call addTacingInfo() here?
```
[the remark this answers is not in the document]
blitz-1306: Fixed in https://github.com/ConsenSys/scribble/pull/180/commits/c5e0537b6f1900ecd339ed2173c5000b362e6bd8

src/spec-lang/expr_grammar.pegjs
```diff
@@ -193,6 +194,17 @@ TypedArgs =
         );
     }
 
+UserConstantDefinition =
+    type: CONST __ md: AnnotationMD? __ const_type: Type __ name: Identifier __ ":=" __ value: Expression {
```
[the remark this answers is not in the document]
blitz-1306: Fixed in https://github.com/ConsenSys/scribble/pull/180/commits/c5e0537b6f1900ecd339ed2173c5000b362e6bd8

src/spec-lang/expr_grammar.pegjs
```diff
@@ -193,6 +194,17 @@ TypedArgs =
...
+            name,
+            const_type,
+            value,
+            md === null ? undefined : md,
+            makeRange(location(), options as ParseOptions)
+        );
+    }
+
 UserFunctionDefinition =
     type: DEFINE __ md: AnnotationMD? __ name: Identifier __ "(" __ args: TypedArgs? __ ")" __ returnType: Type __ "=" __ body: Expression {
```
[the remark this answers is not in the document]
blitz-1306: Fixed in https://github.com/ConsenSys/scribble/pull/180/commits/c5e0537b6f1900ecd339ed2173c5000b362e6bd8

src/spec-lang/tc/typeenv.ts
```diff
@@ -1,24 +1,101 @@
...
+
+        this.mapping.set(scope, defs);
+    }
+
+    pp(): string {
+        return "<user_constants_scope>";
+    }
+}
+
+export class UserFunctionsScope {
```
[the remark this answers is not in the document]
blitz-1306: Fixed in https://github.com/ConsenSys/scribble/pull/180/commits/c5e0537b6f1900ecd339ed2173c5000b362e6bd8

test/samples/consts_simple.instrumented.sol
blitz-1306: I would say yes, except if `sortContracts()` during `tcUnits()` will be involved somehow. But I did discovered nay cases when this actually happends... Declating base contract _after_ inheriting in single file is restricted by compiler since 0.4.13. The following source can not even be compiled (tested in Remix IDE):

```solidity
/// #const bool W := D * 7;
contract Other is Some {
    /// #if_succeeds W == D * 7;
    function testWHD() public {}
}

/// #const uint256 H := 60 * 60;
/// #const uint256 D := H * 24;
contract Some {
    /// #if_succeeds D == H * 24;
    function testHD() public {}
}
```

blitz-1306: Except the inheritance, Scribble do not have any other reasons to change processing order of contracts.

test/multifile_samples/node_modules_erc20/contracts/Foo.sol.instrumented
blitz-1306: Removed in https://github.com/ConsenSys/scribble/pull/180/commits/c5e0537b6f1900ecd339ed2173c5000b362e6bd8

src/instrumenter/instrument.ts
```diff
@@ -670,6 +679,69 @@ function makeUserFunctions(
...
+    }
+
+    const factory = ctx.factory;
+    const nameGen = ctx.nameGenerator;
+
+    /**
+     * @todo This is a hack as we need function scope to transpile annotations.
+     * Consider other ways in the future to do this;
+     */
+    const stubFunc = factory.addEmptyFun(
```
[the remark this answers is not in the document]
blitz-1306: Added to common test sample in https://github.com/ConsenSys/scribble/pull/180/commits/315a6fea533b94db8bc984accccbde30087213c5

test/samples/consts_simple.instrumented.sol
cd1m0: This makes sense. I think the logic is ok then

src/spec-lang/tc/typeenv.ts
```diff
@@ -1,24 +1,98 @@
...
+
+/**
+ * Class to map user-defined constants to their scoping areas
+ */
+export class UserConstantScoping extends UserDefinitionScoping<
+    SourceUnit,
+    SUserConstantDefinition
+> {
+    get(scope: SourceUnit, name: string): SUserConstantDefinition | undefined {
+        const defs = this.mapping.get(scope);
```
cd1m0, line 39: I wonder if we should allow constants defined in base classes to be used in the child classes? In which case the code for this would be the same as the UserFunctionScoping? Lets discuss offline

cd1m0 (changes requested):
Awesome work! Very few nits left, and one thing to discuss. Oh also there is a small change needed in `scId` in `semcheck.ts`. We should add an `if-` case that looks something like this:

```
} else if (def instanceof SUserConstantDefinition) {
    isConst = true;
} ...
```

src/spec-lang/tc/typecheck.ts
```diff
@@ -979,6 +1012,10 @@ function tcIdVariable(expr: SId, ctx: STypingCtx, typeEnv: TypeEnv): TypeNode |
         return def.iteratorType;
     }
 
+    if (def instanceof SUserConstantDefinition) {
+        return def.formalType;
+    }
+
```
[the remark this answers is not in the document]
blitz-1306: Fixed in https://github.com/ConsenSys/scribble/pull/180/commits/d2914d736fc9029eba8736f27bacc34050199d57

src/instrumenter/instrument.ts
```diff
@@ -670,6 +679,69 @@ function makeUserFunctions(
...
+                "=",
+                factory.makeIdentifierFor(userConst),
+                constValue
+            ),
+            `Value assignment for ${constDef.pp()}`
+        );
+
+        ctorBody.appendChild(stmtAssignValue);
+
+        ctx.addGeneralInstrumentation(userConst, stmtAssignValue);
```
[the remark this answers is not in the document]
blitz-1306: Uanble to address this: if I do this, then `src2src` test is failing. User functions are treated as **general instrumentation** and I'm not yet sure if user constants should be annotations. Maybe I'm missing something here...

Anyway we also skip user funcs and user consts in JSON output, as they are not properties (not instance of `PropertyMetaData`). Maybe there should be some alterations here, but I'm unsure how to implement them properly.

cd1m0 (approved):
LGTM

src/bin/scribble.ts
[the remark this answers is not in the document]
cd1m0: Should be addressed in https://github.com/ConsenSys/scribble/pull/180/commits/6a49ac80f8d95f333b842793128fd91161c224a6.

src/instrumenter/instrument.ts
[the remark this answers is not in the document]
cd1m0: Makes sense.

src/spec-lang/tc/typeenv.ts
cd1m0: Fixed in https://github.com/ConsenSys/scribble/pull/180/commits/6a49ac80f8d95f333b842793128fd91161c224a6 to allow constants in base contracts to be used in inheriting contracts. Also changed the behavior so that constants are only per contract such that we're consistent with user functions.

src/spec-lang/tc/semcheck.ts
```diff
@@ -125,7 +125,16 @@ export function scAnnotation(
         }
         sc(node.expression, ctx, typings, semMap);
     } else if (node instanceof SUserConstantDefinition) {
-        sc(node.value, ctx, typings, semMap);
+        const info = sc(node.value, ctx, typings, semMap);
+
+        if (!info.isConst) {
+            throw new SemError(
+                `Cannot use non-constant expression ${node.value.pp()} in constant definition for ${
+                    node.name
```
blitz-1306, line 133: Shouldn't here be `node.name.name` or `node.name.pp()` instead of `node.name` (which is probably an `SId`)?

src/spec-lang/tc/typeenv.ts
```diff
@@ -91,8 +66,8 @@ export class TypeEnv {
     constructor(compilerVersion: string, abiEncoderVersion: ABIEncoderVersion) {
         this.typeMap = new Map();
 
-        this.userFunctions = new UserFunctionScoping();
-        this.userConstants = new UserConstantScoping();
+        this.userFunctions = new UserDefinitionScoping<SUserFunctionDefinition>();
+        this.userConstants = new UserDefinitionScoping<SUserConstantDefinition>();
```
blitz-1306, line 70: Should these be `UserFunctionScoping` and `UserConstantScoping` instead? We have declared class type aliases but not using them, otherwise.

blitz-1306 (commented):
Looks good. A few notices, though.

src/spec-lang/tc/typeenv.ts
cd1m0: Whenever I try to use those instead I get a "type used as value" error.

src/spec-lang/tc/semcheck.ts
cd1m0: Fixed in [1f5c2ca](https://github.com/ConsenSys/scribble/pull/180/commits/1f5c2cafd2d5c6dce1b8dad70df3ae3515e1abad)

cd1m0 (approved):
LGTM

cd1m0:
There is currently some issue with CI we are still investigating. Tests pass locally so force-merging.

PR #185, opened 2022-07-25
Move CI from Drone to GitHub Actions

blitz-1306:
## Changes
- [x] Removed Drone config and created GitHub Actions config instead.
- [x] Set quality gate in `.nycrc` file to match current coverage threshold.
- [x] Tweaked `package.json`: updated dependencies, tweaked commands to use `lcov` coverage report format, removed obsolete codecov uploader dependency.

## Notes
- CI is still taking 1h per run and that is not fortunate. We should investigate further the ways to either optimize our tests or apply tweaks to get performance boost.
- This PR uses latest version of **solc-typed-ast** for the current moment. Some snapshots and artifacts are updated due to this.
- Additional removal of Drone WebHooks in repository settings is also needed.

Regards.


cd1m0 (approved):
LGTM

PR #203, opened 2022-11-18
Drop builtin type duplicates

blitz-1306:
## Preface
This PR extends #200, but might be slightly more unstable than #200 currently. It drops Scribble's own builtin representations and relies on same components from **solc-typed-ast** instead. So I'm creating a standalone PR to leave a decision if we want to merge #200 and release Scribble and do more testing of changes in this branch.

## Changes
- [x] Remove duplicated builtin declarations.
- [x] Tune up related logic and tests.

## Concerns
There are several TODOs that worth considering.
- One is related to `FunctionSetType`, that is slightly different than one in **solc-typed-ast**.
- One is related to processing builtin functions in `scFunctionCall()`. Builtin function types in **solc-typed-ast** does not have state mutability indicator, that was used to check if callable function can be considered as constant expression. I was forced to turn this logic off for builtins and consider expressions with them as non-constant. We can decide if we want to intorduce state mutability to **solc-typed-ast**'s builtins and then restore this logic. Unsure about implications of current changes. However  tests are green with current changes and maybe I'm just too cautious. Would like to get feedback here.

Regards.


src/spec-lang/tc/semcheck.ts
```diff
@@ -532,9 +533,19 @@ export function scFunctionCall(
...
         const isConst = calleeT.mutability === FunctionStateMutability.Pure && allArgsConst;
+
         return { isOld: ctx.isOld, isConst: isConst, canFail: true };
     }
 
+    /**
+     * @todo This is not precise as it was with typeString parser
+     * due to not all builtins have state mutability indicator.
+     * Figure out what to do with it.
+     */
```
cd1m0, line 544: Its ok to be a bit more imprecise here.

src/spec-lang/tc/typecheck.ts
```diff
@@ -1539,30 +1493,29 @@ export function tcIndexAccess(expr: SIndexAccess, ctx: STypingCtx, typeEnv: Type
...
 
-        if (type === undefined) {
-            throw new SInaccessibleForVersion(expr, expr.member, typeEnv.compilerVersion);
-        }
+    if (baseT instanceof BuiltinStructType) {
+        return getFieldOfBuiltinStruct(baseT, expr, typeEnv);
+    }
 
-        return type;
+    if (baseT instanceof BuiltinFunctionType && baseT.returns[0] instanceof BuiltinStructType) {
```
cd1m0, line 1517: I am confused by this logic. `baseT` is a function type. So the base refers to the function itself, not to a function call. So even if `baseT` returns some builtin struct, we shouldn't be able to resolve fields on `baseT` right? Only if `baseT` was called.

cd1m0 (changes requested):
Good job! One question only

src/spec-lang/tc/typecheck.ts
blitz-1306: I guess this is edge-case for `type()` or so, that returns a builtin struct. So this is more like an edge case. Maybe there is better way of addressing it.

cd1m0: The fix was to change how we type the `type` identifier in [021aa55](https://github.com/ConsenSys/scribble/pull/203/commits/021aa552739d19bc244aa72209ac5225f40f4d2c)

cd1m0 (approved):

PR #204, opened 2022-12-06
Upgrade type inference dependent components to respect changes with ABI encoder

blitz-1306:
## Changes
- [x] Updated dependencies (use [**solc-typed-ast** v12.0.0](https://github.com/ConsenSys/solc-typed-ast/releases/tag/v12.0.0)).
- [x] Tweaked type inference dependent components to respect changes with ABI encoder.
- [x] Adjusted `.nycrc` to skip parser header files.
- [x] Introduced `--download-compilers` CLI option and related test. Simplified GitHub Actions flow and updated README.
- [x] Tweaked affected tests.

Regards.


blitz-1306:
Current notes and concerns due to self-review:
- It seems that getting `SourceUnit` user-defined ABI encoder version can be considered as a getter. There might be better to skip involving `compilerVersion` into the process and apply it somewhere later (depending on circumstances, when there is no user-defined ABI encoder version).
- Making multiple instances of `InferType` will significantly affect code base. So maybe it is worth to figure out encoder version for all of the units to build single type inference instance. Currently a very hacky implementation is in `getABIEncoderVersionForUnits()`. If such logic is universal, then it is better to be placed in **solc-typed-ast** for convenience.
- There is a case when current implementation may crash:
    ```solidity
    // a.sol
    pragma solidity ^0.8.0;
    pragma abicoder v2;

    contract Some {}
    ```
    ```solidity
    // b.sol
    pragma solidity ^0.8.0;
    pragma abicoder v1;

    import "./a.sol";

    contract Other is Some {}
    ```

cd1m0:
After discussion with @blitz-1306 it seems that:

1. Its possible for a V1 contract to inherit from a V2 contract. Consider these samples:

**a.sol**

```import "./b.sol";
pragma abicoder v2;

contract A is B {
	struct Sa {
		uint t;
	}

	Sa public x;
}
```

**b.sol**

```
pragma abicoder v1;

contract B {

	struct Sb {
		address t;
		bool d;
	}

	Sb public y;
}
```

Compiling  **a.sol** succeeds: `sol-ast-compile a.sol --compiler-version 0.8.17 `. We can confirm that `B` is **indeed** compiled with V1 of the abi, since if we add a nested struct to `B` it will fail compiling:

```
pragma abicoder v1;

contract B {
	struct Sb1 {
		uint t;
	}
	struct Sb {
		address t;
		bool d;
		Sb1 s;
	}

	Sb public y;
}
```

2. Note that it is **also** possible for a V1 contract to inherit from a V2 contract, as long as there are no features in the V2 contract that **need** the V2 encoder. For the below samples, compiling `a1.sol` **also succeeds**:

**a1.sol**
```
import "./b1.sol";
pragma abicoder v1;

contract A is B {
}
```

**b1.sol**
```
pragma abicoder v2;

contract B {
	struct S {
		uint t;
	}

	S public s;
}
```

3. However if the V2 base contract **needs** a V2 encoder (like if it encodes nested structs), then we fail. In the below 2 samples, compiling `a2.sol` will fail:

**a2.sol**
```
import "./b2.sol";
pragma abicoder v1;

contract A is B {
}
```

**b2.sol**
```
pragma abicoder v2;

contract B {
	struct S1 {
		address a;
	}

	struct S {
		uint t;
		S1 s;
	}

	S public s;
}
```

blitz-1306:
Tok some time to look through compiler implementation. Found following for a reference:

- Basically check calls of following two functions in `TypeChecker` plus `ContractLevelChecker` and it would show edge-cases
  - https://github.com/ethereum/solidity/blob/1c8745c54a239d20b6fb0f79a8bd2628d779b27e/libsolidity/analysis/TypeChecker.cpp#L59-L73
  - https://github.com/ethereum/solidity/blob/1c8745c54a239d20b6fb0f79a8bd2628d779b27e/libsolidity/analysis/TypeChecker.cpp#L3968-L3975
  - https://github.com/ethereum/solidity/blob/1c8745c54a239d20b6fb0f79a8bd2628d779b27e/libsolidity/analysis/ContractLevelChecker.cpp#L482-L528

Also, some of the related test samples:
- https://github.com/ethereum/solidity/blob/1c8745c54a239d20b6fb0f79a8bd2628d779b27e/test/libsolidity/syntaxTests/imports/inheritance_abi_encoder_match.sol
- https://github.com/ethereum/solidity/blob/1c8745c54a239d20b6fb0f79a8bd2628d779b27e/test/libsolidity/syntaxTests/imports/inheritance_abi_encoder_mismatch_1.sol
- https://github.com/ethereum/solidity/blob/1c8745c54a239d20b6fb0f79a8bd2628d779b27e/test/libsolidity/syntaxTests/imports/inheritance_abi_encoder_mismatch_2.sol
- https://github.com/ethereum/solidity/blob/1c8745c54a239d20b6fb0f79a8bd2628d779b27e/test/libsolidity/syntaxTests/abiEncoder/v1_inheritance_from_contract_calling_v2_function.sol
- https://github.com/ethereum/solidity/blob/1c8745c54a239d20b6fb0f79a8bd2628d779b27e/test/libsolidity/syntaxTests/abiEncoder/v1_inheritance_from_contract_defining_v2_function_accepting_struct.sol
- https://github.com/ethereum/solidity/blob/1c8745c54a239d20b6fb0f79a8bd2628d779b27e/test/libsolidity/syntaxTests/abiEncoder/v1_inheritance_from_contract_defining_v2_function_returning_struct.sol

blitz-1306:
Also, all of the compiler checks are executed here (note the order): https://github.com/ethereum/solidity/blob/1c8745c54a239d20b6fb0f79a8bd2628d779b27e/libsolidity/interface/CompilerStack.cpp#L483-L503

cd1m0 (approved):

PR #210, opened 2023-02-21
Issue.202

cd1m0:
This PR changes the instrumentation to:

1. Create a new `__ScribbleUtilsLib__` library, for each instrumented file that contains the definitions for the `AssertFailed` and `AssertFailedData` fields. Also helper functions `isInContract` and `setInContract` that implement the old 'out-of-contract' flag functionality from the scribble utils base class.

2. Move the array sum functions to the `__ScribbleUtilsLib__`

3. Move the user constants from the ScribbleUtils base class to the contract where they are defined.

4. Remove the ScribbleUtils base class

5. Remove the needs to import the scribble utils imports


blitz-1306 (approved):
Looks good to me

PR #211, opened 2023-02-28
Add the instrumetned source location to error strings

cd1m0:
This PR changes the error string that are emitted in the `AssertionFailed` events to include the source triple corresponding to the location in the instrumented file, where the event is thrown. So for example instrumented code that used to look like this:

```
        if (!(y == (x + 1))) {
            emit __ScribbleUtilsLib__16.AssertionFailed("0: P0");
            assert(false);
        }
```

Would now look like this:

```
        if (!(y == (x + 1))) {
            emit __ScribbleUtilsLib__16.AssertionFailed("000296:0068:000 0: P0");
            assert(false);
        }
```

Where the string starting at offset 296 with length 68 would be precisely the emit statement above.

Implementing this is a little tricky, as we don't know the exact offset of a statement in the instrumented file until we generate the instrumented file. So to implement this we have a slightly hacky 2-pass solution:

1. Generate the instrumented AST with event string literals that include a placeholder string `"000000:0000:000"`, record in the `InstrumentationContext` to which AST node's location should each string literal be fixed (see `InstrumentationContext.addStringLiteralToAdjust`)
2. Generate the instrumented files
3. Find where the literals are emitted in the instrumented files, and fix-them-up in-place using the mapping in the `InstrumentationContext` (see `InstrumentationContext.adjustStringLiterals`)


src/instrumenter/instrumentation_context.ts
```diff
@@ -467,6 +471,52 @@ export class InstrumentationContext {
...
+                contents.slice(strLoc[0] + 1, strLoc[0] + 16)
+            );
+
+            contents =
+                contents.slice(0, strLoc[0] + 1) +
+                `${String(newLoc[0]).padStart(6, "0")}:${String(newLoc[1]).padStart(
+                    4,
+                    "0"
+                )}:${String(fileInd).padStart(3, "0")}` +
+                contents.slice(strLoc[0] + 16);
```
blitz-1306, line 514: This looks slightly complex to read: a lot of formatting is happening in-place. Would it be possible to compose source location in variable before updating `contents`? It would simplify update assignment.

test/integration/scribble/run/valid.spec.ts
```diff
@@ -40,7 +40,7 @@ describe(`Command "scribble <filename>" is working properly`, () => {
 
             it("STDOUT is correct", () => {
                 // Uncomment next line to update instrumented sources
-                // fse.writeFileSync(instrFileName, output, { encoding: "utf-8" });
+                fse.writeFileSync(instrFileName, output, { encoding: "utf-8" });
```
blitz-1306, line 43: I guess this should be commented out.

src/instrumenter/instrument.ts
```diff
@@ -228,10 +228,11 @@ function emitAssert(
     let userAssertionHit: Statement | undefined;
 
     if (instrCtx.assertionMode === "log") {
-        const strMessage = `${annotation.id}: ${annotation.message}`;
+        const strMessage = `000000:0000:000 ${annotation.id}: ${annotation.message}`;
```
blitz-1306, line 231: A question out of curiosity: is there a particular reason to use padded for zeroes for initial value?

src/instrumenter/instrumentation_context.ts
```diff
@@ -467,6 +471,52 @@ export class InstrumentationContext {
...
+        units: SourceUnit,
+        fileInd: number,
+        srcMap: SrcRangeMap
+    ): string {
+        for (const lit of units.getChildrenByType(Literal)) {
+            if (!(lit.kind === LiteralKind.String && this.litAdjustMap.has(lit))) {
+                continue;
+            }
+
+            const targetNode = this.litAdjustMap.get(lit) as ASTNode;
```
blitz-1306, line 489: We traversing all literals in source unit here and skipping ones that are not in `this.litAdjustMap`. Seem to me that we can use following shortcut more effectively:
```typescript
for (const [lit, targetNode] of this.litAdjustMap) {
   // ...
}
```

blitz-1306 (commented):
Good job. Few questions and suggestions.

src/instrumenter/instrumentation_context.ts
blitz-1306: P.S.: Also `units` is actually a single source unit.

src/instrumenter/instrumentation_context.ts
cd1m0: Fixed in [029c9c3](https://github.com/ConsenSys/scribble/pull/211/commits/029c9c3117655fdbc0f0f348bcd11e354e5ce2ad)

test/integration/scribble/run/valid.spec.ts
cd1m0: Fixed in [029c9c3](https://github.com/ConsenSys/scribble/pull/211/commits/029c9c3117655fdbc0f0f348bcd11e354e5ce2ad)

src/instrumenter/instrument.ts
cd1m0: Just as a way to remember how the source locations should be formatted. It could have just as easily be empty spaces, just made sense in my head

src/instrumenter/instrumentation_context.ts
cd1m0: Fixed in [029c9c3](https://github.com/ConsenSys/scribble/pull/211/commits/029c9c3117655fdbc0f0f348bcd11e354e5ce2ad)

blitz-1306 (approved):

PR #212, opened 2023-03-02
Support Solidity 0.8.19

blitz-1306:
## Changes
- [x] Updated dependencies.
- [x] Updated affected test artifacts. Add test sample to use operators for custom types.
- [x] Tweak logic to handle tuple types with empty elements.
- [x] Tweak call graph to respect operators for custom types.

Regards.


src/instrumenter/callgraph.ts
```diff
@@ -33,6 +35,25 @@ export interface CallGraph {
     overridenBy: FunMap;
 }
 
+function findCallersAndCallees(fun: FunctionDefinition, callers: FunMap, callees: FunMap): void {
```
cd1m0, line 38: Can we rename this function to `processCallNodes` ?

cd1m0 (changes requested):
LGTM. I need to do a pass and add TC logic to support custom operators in annotations

src/instrumenter/callgraph.ts
blitz-1306: Sure, will do.

cd1m0 (approved):
LGTM

PR #213, opened 2023-03-06
Fix nested macro variable substitution in complex SNodes

blitz-1306:
This PR fixes children enumeration for several `SNode` descendants, such as `SFunctionCall`. This fixes an issue, that not all macro variable placeholders were properly substituted during `SNode` children traversal.

Also this PR adds some more cached snapshots to slightly boost tests.


cd1m0 (dismissed):
LGTM

cd1m0 (dismissed):
lgtm

Issue #122, opened 2021-11-18
Support structural equality comparison of non-value types in Scribble


cd1m0:
We have use cases for comparing strings, arrays and structs in properties. However the base Solidity language doesn't support equality comparison of non-value types.

We could add structural equality comparison support to Scribble, by transpiling it into comparison between the `abi.encode`-ed versions of the data structures. So for example this:

```solidity
contract Foo {
   /// #if_succeeds a == b;
   function main(string memory a, string calldata b) public { ... }
}
```

Would be transpiled to:

```solidity
contract Foo is __scribble_ReentrancyUtils  {
    event AssertionFailed(string message);

    function main(string memory a, string calldata b) public {
        _original_Foo_main(a, b);
        unchecked {
            if (!(bytesEq(abi.encode(a), abi.encode(b)))) {
                emit AssertionFailed("0: ");
                assert(false);
            }
        }
    }

    function _original_Foo_main(string memory a, string calldata b) private { ... }
}
/// Utility contract holding a stack counter
contract __scribble_ReentrancyUtils {
    bool __scribble_out_of_contract = true;
    
    function bytesEq(bytes memory a, bytes memory b) internal returns (boolean) {
      if (a.length != b.length) { return false; }
      
      for (uint i = 0; i < a.length; i++) {
        if (a[i] != b[i]) return false;
      }
      
      return true;
    }
}
```

wuestholz:
@cd1m0 Maybe we should use a different operator (e.g., `encodingEq(a, b)`) instead of `==` to avoid confusion. What do you think?

In general, it seems like this can become very expensive in terms of gas and we might want to be careful with supporting it. Hashing the bytes might be cheaper. Maybe `bytesEq` could have a separate early-return if the hashes are not equal.

blitz-1306:
I started implementation in #224 to have some PoC. There are several questions to address:

1. Do we want to extend `==`, use custom operator (like `=^=` for example), or provide builtin for this functionality?
2. Do we want to apply `abi.encode()` / `abi.encodePacked()` for the user (slightly rewrite code on call sites)?
3. Do we want to refactor builtins addition flow to Scribble (as in current state it requires minor chages in multiple places to support new builtin function) or leave it to a followup pass?

Let me know what you think. Thanks.

blitz-1306:
This is implemented in #224 and would be a part of next release.

PR #224, opened 2023-05-10
Provide mechanic to check equality of non-value types

blitz-1306:
## Status
This is PR a PoC implementation for #122.

## Changes
- [x] Introduce `eq_encoded()` Scribble builtin to simplify comparing of complex types.
  - Function implementaion is added to Scribble utils contract on-demand (when function is used by any annotation).
  - When any passed argument is not castable to `bytes memory`, it is wrapped with `abi.encode()` (to reduce amount of manual efforts).
- [x] Optimize code of generating utils library (introduce more granular functions).
- [x] Add test sample.

## Notes
Feature discussion goes in #122. See some questions in https://github.com/ConsenSys/scribble/issues/122#issuecomment-1541514034.

Regards.


src/spec-lang/tc/typecheck.ts
```diff
@@ -1962,97 +1962,124 @@ export function tcForAll(expr: SForAll, ctx: STypingCtx, typeEnv: TypeEnv): Type
...
+                    expr
+                );
+            }
+
+            /**
+             * @todo Consider checking that type is encodable.
+             * Here is how compiler does it:
+             * https://github.com/ethereum/solidity/blob/0a0c389541c0f247691951edd64451be7145436c/libsolidity/ast/Types.cpp#L314-L339
+             */
+            tc(expr.args[0], ctx, typeEnv);
```
cd1m0, line 2020: Here we need to check that both argument types are 'abi-encodable'. Maybe we can add a new function `isTypeABIEncodable(t: TypeNode): bool` that follows the logic of `toABIEncdodedType`

test/samples/eq_encoded.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
```
cd1m0, line 2: Could we add an executable regression test that calls eq_encoded (i.e. a vm test?)?

We should have a negative case (checks that eq_encoded fails when we want it to), and a positive test, for each of the following cases:

1. Comparing two things of different types (e.g. number and array) - negative test, always fails
2. Comparing two primitive things of the same type - numbers, addresses, booleans, fixed bytes, enums (should succeed when equal, fail when not)
3. Comparing two arrays (again check that eq_encoded succeeds when equal, fails when not)
4. Comparing two structs (same 2 tests)

cd1m0 (changes requested):
Good job! Some small tweaks and tests requested

test/samples/eq_encoded.sol
blitz-1306: Will do.

src/spec-lang/tc/semcheck.ts
```diff
@@ -481,6 +481,16 @@ export function scFunctionCall(
             return { isOld: isOld, isConst: allArgsConst, canFail: argsInfo[0].canFail };
         }
 
+        if (callee.name === ScribbleBuiltinFunctions.eq_encoded) {
+            const isOld = ctx.isOld || argsInfo.some((argInfo) => argInfo.isOld);
```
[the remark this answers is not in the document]
blitz-1306: Will address.

src/instrumenter/utils_library.ts
```diff
@@ -320,15 +305,214 @@ export function generateUtilsLibrary(
...
+        false,
+        DataLocation.Default,
+        StateVariableVisibility.Default,
+        Mutability.Mutable,
+        "<missing>",
+        undefined,
+        factory.makeElementaryTypeName("uint256", "uint256")
+    );
+
+    const byteCheck = factory.makeForStatement(
```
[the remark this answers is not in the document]
blitz-1306: Will do.

test/samples/eq_encoded.sol
blitz-1306: Addressed in e060422

src/spec-lang/tc/semcheck.ts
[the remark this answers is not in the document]
blitz-1306: Addressed in e060422

src/instrumenter/utils_library.ts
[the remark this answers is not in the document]
blitz-1306: Addressed in e060422

src/spec-lang/tc/typecheck.ts
blitz-1306: See https://github.com/ConsenSys/solc-typed-ast/issues/206

cd1m0 (approved):
lgtm

PR #225, opened 2023-05-15
ci: notify after build step on new merges to master

psantos-consensys:


.github/workflows/node.js.yaml
```diff
@@ -1,57 +1,85 @@
...
+        with:
+          name: codecov-umbrella
+          directory: ./coverage/
+          fail_ci_if_error: true
+          verbose: true
+
+  notify-slack:
+    needs: build
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
```
blitz-1306, line 62: Can we also extend it to inform about merges to `develop`? We use two branches in this repository.

cd1m0 (changes requested):
@psantos-consensys could you just address Pavel's comment below? Then it will be ready to merge

.github/workflows/node.js.yaml
psantos-consensys: removed the if statement and modified the env to pick the branch name

cd1m0 (dismissed):
lgtm

blitz-1306 (approved):

blitz-1306 (approved):

PR #229, opened 2023-05-24
Support Solidity 0.8.20

blitz-1306:
## Notes
This PR adds support of Solidity 0.8.20 to Scribble. It is based on not yet released version of **solc-typed-ast** that is tested locally. Main goals are following:
- Make an end-to-end test of pending changes in **solc-tyed-ast**.
- Check how Scribble will adapt to changes and consider implications.
- Have a proper discussion.

## Changes
- [x] Use unreleased **solc-typed-ast** with Solidity 0.8.20 support (see https://github.com/ConsenSys/solc-typed-ast/pull/208).
- [x] Tweak codebase where we use constructors of `ContractDefinition` due to their signature change.
- [x] Use `InferType.isABIEncodable()` for type-checking `eq_encodable()` args.
- [x] Tweak tests and snapshots.

## Notes
- EVM version in latest Solc is now `Shanghai`, so there was changes in VM tests, related to sepecting proper VM version for different requirements of Solidity versions in sources.
- Enums are now written differenly by solc-typed-ast writers, so some sourcemaps changed. I verified a few and things seem to be fine (even in case of dangling docs).

Regards.


cd1m0 (approved):
lgtm

PR #235, opened 2023-07-19
Introduce "hardhat" assertion mode

### LINKED ISSUE #114: Support new 'hardhat' `--user-assert-mode`
Scribble is sometimes used to instrument the code for local testing. When this is done in a Hardhat environment, people often use `console.log(...)` to print out various information. In this mode, it is more convenient to signal user assertion failures with `console.log()` instead of emitting events, so that users can just look at their terminal instead of go hunting through the logs. In this mode we would support a new option for the `--user-assert-mode` key as follows.

Given the below sample:

```
contract Foo {
        /// #if_succeeds {:msg "Property 0 fail"} x > 0;
        function foo(uint x) public {}
}
```

When we instrument it with `scribble --user-assert-mode hardhat tmp.sol` we would get the following code:

```
import "hardhat/console.sol";

contract Foo {
    function foo(uint x) public {
        _original_Foo_foo(x);
        if (!(x > 0)) {
            console.log("0: Property 0 fail");
            assert(false);
        }
    }

    function _original_Foo_foo(uint x) private {}
}
```

Debug events would also be converted to `console.log`s. So running  `scribble --user-assert-mode hardhat tmp.sol --debug-events` would produce:

```
import "hardhat/console.sol";

contract Foo {
    function foo(uint x) public {
        _original_Foo_foo(x);
        if (!(x > 0)) {
            console.log("0: Property 0 fail");
            console.log("x:", x);
            assert(false);
        }
    }

    function _original_Foo_foo(uint x) private {}
}
```

Note that the `console.log()` functions in hardhat are limited to up to 4 arguments, so extra work needs to be done when we need to print more than 4 values.

blitz-1306:
## Preface
This PR resolves #114.

## Changes
- [x] Introduce `hardhat` assertion mode.
  - Remove unused events from Scribble utils library for `hardhat` assertion mode.
  - Tweaked logic to respect new mode when emitting assertion failures.
- [x] Introduce adding `import "hardhat/console.sol";` in instrumented source units (flat and non-flat).
- [x] Tuned tests to verify the correct behavior for instrumented sources in new mode.

## Notes
- See `console.log()` limitations, mentioned here: https://hardhat.org/hardhat-network/docs/reference#console.log
- See `console.log()` implementation here: https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/console.sol

Regards.


src/instrumenter/instrument.ts
```diff
@@ -154,22 +153,32 @@ function getDebugInfoEmits(
...
-                "<missing>",
-                FunctionCallKind.FunctionCall,
-                factory.makeIdentifier("<missing>", "abi.encode", -1),
-                evtArgs
-            )
-        ]);
+        if (instrCtx.assertionMode === "hardhat") {
+            const emitStmts: Statement[] = [];
+
+            for (const evtArg of evtArgs) {
```
cd1m0, line 160: Why are we logging the arguments one by one? Can't we emit a single log statement with all the event args (and maybe the event location/name?

cd1m0 (changes requested):
Looks like a good start. One question

src/instrumenter/instrument.ts
blitz-1306: In #114 there is an entry:

> Note that the `console.log()` functions in hardhat are limited to up to 4 arguments, so extra work needs to be done when we need to print more than 4 values.

It seems cheaper to implement standalone logging. Otherwise we whould have to chunk-split the event args and emit `console.log()` per each pair. I'm also not sure if event args are always instance of `Identifier` to also write their names. WDYT?

blitz-1306: Partially fixed in 4e6e130

blitz-1306: Again, removed "chunking" logic due to poor support of `bytes memory` in overloaded `console.log()`. It would be a significant logic compilcation to decide if Scribble should use over `console.log()` that supports very limited se of primitive args, or specific `console.log*()` function for standalone argument. Give that there is a possibility to hit compliex types and wrap them around with `abi.encode()`, we better stick to single solution. See bf8a65a.

src/instrumenter/instrument.ts
```diff
@@ -243,6 +374,13 @@ function emitAssert(
                 factory.makeLiteral("<missing>", LiteralKind.String, "", `HIT: ${strMessage}`)
             ]);
         }
+    } else if (instrCtx.assertionMode === "hardhat") {
+        const strMessage = `000000:0000:000 ${annotation.id}: ${annotation.message}`;
```
cd1m0, line 378: strMessage is already defined on line 362

src/instrumenter/instrumentation_context.ts
```diff
@@ -375,10 +377,12 @@ export class InstrumentationContext {
 
     private readonly litAdjustMap = new Map<Literal, ASTNode>();
 
+    private readonly unitsNeedsImports = new Map<SourceUnit, string>();
```
cd1m0, line 380: In the future we may use this to say a single SourceUnit needs multiple different imports. Currently the type of this map allows only 1 import path per unit. Could you change this to:
```
    private readonly unitsNeedsImports = new Map<SourceUnit, Set<string>>();
```

cd1m0 (changes requested):
Almost ready. 2 small nits

src/instrumenter/instrument.ts
blitz-1306: Addressed in 10c64ec

src/instrumenter/instrumentation_context.ts
blitz-1306: Addressed in 10c64ec

cd1m0 (approved):
lgtm!

PR #236, opened 2023-08-04
A followup polishing pass

blitz-1306:
## Preface
This PR fixes several issues, that are found during the implementation of #235 (`hardhat` user assertion mode support).

## Changes
- [x] Fix bug in annotation hex values writing to Solidity. Example (that would fail with current logic):
```
/// #if_succeeds let y := bytes1(0xf) in y > 0x1;
```
- [x] Tweak literal adjustment logic to also process `cov-assertion`s literals.
- [x] Add test case to verify hex literals behavior. Fix other test artifacts.
- [x] Rename `scUnits()` to `scAnnotations()` as it does not use source units.

Regards.


cd1m0 (approved):
lgtm

PR #239, opened 2023-09-01
Rewrite immutable state variables to be mutable

### LINKED ISSUE #237: Instrumentation for `immutable`
If state variable annotation present on a contract, `scribble` generates an auxiliary function with the original constructor body and calls it in the the constructor. That newly generated function is preceded/succeeded by `scribble` generated code, according to the annotations. However, if an `immutable` is set in the constructor, the newly generated function tries to set the `immutable` inside its body which is not allowed by solidity. When the contract is being compiled after arming, the following error can be observed:

```
[⠒] Compiling...
[⠑] Compiling 1 files with 0.8.20
[⠘] Solc 0.8.20 finished in 9.39ms
Error: 
Compiler run failed:
Error (1581): Cannot write to immutable here: Immutable variables can only be initialized inline or assigned directly in the constructor.
  --> src/Example.sol:21:9:
   |
21 |         x = 5;
   |         ^
```

Example contract:
```solidity
/// #if_succeeds x == 5;
contract Example {
    uint private immutable x;
    constructor() {
        x = 5;
    }
}
```

After arming:
```solidity
/// #if_succeeds x == 5;
contract Example {
    uint private immutable x;

    constructor() {
        _original_Example_constructor();
        unchecked {
            if (!(x == 5)) {
                emit __ScribbleUtilsLib__14.AssertionFailed("000389:0066:000 0: ");
                assert(false);
            }
        }
    }

    function _original_Example_constructor() private {
        x = 5;
    }
}

library __ScribbleUtilsLib__14 {

    event AssertionFailedData(int eventId, bytes encodingData);

    function assertionFailed(string memory arg_0) internal {
        emit AssertionFailed(arg_0);
    }

    function assertionFailedData(int arg_0, bytes memory arg_1) internal {
        emit AssertionFailedData(arg_0, arg_1);
    }

    function isInContract() internal returns (bool res) {
        assembly {
            res := sload(0x5f0b92cf9616afdee4f4136f66393f1343b027f01be893fa569eb2e2b667a40c)
        }
    }

    function setInContract(bool v) internal {
        assembly {
            sstore(0x5f0b92cf9616afdee4f4136f66393f1343b027f01be893fa569eb2e2b667a40c, v)
        }
    }
}
```

Scribble version used: `0.6.23`

blitz-1306:
## Preface
This PR fixes #237. While solution might not be the best, still it is able to handle specified edge-case and some others.

## Changes
- [x] Rewrite `immutable` state variables to be mutable.
- [x] Introduce related test sample (including few other edge-cases).

## Notes
It may seem debateable if we should interpose constructors, as mentioned in following inline comment:
https://github.com/Consensys/scribble/blob/ccc74a7a7a8dd48b3e915ea14e62272dc9491624/src/bin/scribble.ts#L400

Still, if someone would like to tune up function filter for interposing, then following code block should be adjusted: https://github.com/Consensys/scribble/blob/ccc74a7a7a8dd48b3e915ea14e62272dc9491624/src/bin/scribble.ts#L385-L411

I'm not changing it due to it would likely cause a lot of implications and would lead to compatibility breaks.

Regards.


wuestholz (approved):
LGTM

Issue #241, opened 2023-09-12
Deal with constructor interposing logic


blitz-1306:
Recently we introduced a fix #239 for #237. However, this is a workaround that we can roll back if we deal with cases when user annotates constructor, and constructor gets interposed. We can introduce logic to differ constructors and other functions to process them  in such a way to not hit edge-cases with immutable state variables. 

There are relevant comments here: https://github.com/Consensys/scribble/issues/237#issuecomment-1698540173, https://github.com/Consensys/scribble/issues/237#issuecomment-1700422333

Please, consider.

PR #243, opened 2023-09-16
Fix for #241

cd1m0:
This PR:

- provides a fix for #241 - removes interposing on constructors (instead annotations are compiled inline in the constructor), rolls-back the re-writing of immutable state vars to mutable;
- fixes a crash in type checking for certain cases of state vars with annotations and a using-for pragma
 
Note:

After this fix we no longer support `#try` and `#require` annotations on constructors, that mention immutable state variables. We believe that this is a rare enough case that it shouldn't impact users.


blitz-1306 (approved):
Good work. Lets see if compiler would pass 0.8.20. Otherwise I fix tests to respect 0.8.21 version (that is recent).

blitz-1306 (dismissed):

blitz-1306 (approved):

PR #248, opened 2023-10-03
Dont interpose on callsites in view functions

cd1m0:
We currently interpose on callsites inside view functions. The original reason for this is that we were worried about external view functions calling back into the contract and potentially getting inconsistent state. Specifically the scenario we were worried about was:

1. We are  in some non-view function F1 of contract C, that violates the invariant
2.  F1 calls F2 (internally), and F2 is a view function
3.  F2 calls (externally) F3 which is also view
4.  F3 calls back into C, and gets inconsistent state
5.  Inconsistent state gets propagated all the way back in to F1

However even if this happened, since the calls to F3 (and onwards) are all view, (and thus side-effect free), then the only thing that the inconsistent state of C can get propagate to is back to C itself. So logically this is similar to just doing internal calls to C (and potentially getting some extra consistent state from external contracts)

For this reason we decided not to instrument external callsites inside view functions, since these can sometimes break hardhat tests.

### DIFF
--- a/src/instrumenter/instrument.ts
+++ b/src/instrumenter/instrument.ts
@@ -1054,9 +1054,18 @@ function replaceExternalCallSites(
             callSite
         );
 
+        // We skip instrumenting external calls in pure and view functions to
+        // avoid changing their mutability. This should not result in missing
+        // any invariant violations since the external call can only call
+        // another view/pure function, and as such can only return information
+        // to the current context without any other side effects. Thus even if
+        // there is a callback to this contract, that sees inconsistent state,
+        // its still functionally equivalent to an internal call chain since
+        // there are no other side-effects outside of this contract.
         if (
             calleeType instanceof FunctionType &&
-            calleeType.mutability === FunctionStateMutability.Pure
+            (calleeType.mutability === FunctionStateMutability.Pure ||
+                calleeType.mutability === FunctionStateMutability.View)
         ) {
             continue;
         }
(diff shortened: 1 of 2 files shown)


blitz-1306 (approved):
Approve this. Still, there maybe some inconsistent behavior in case of older Solidity (`0.4`) that may allow calling changing state function fron non-chaning state function. Effects of this case are uncertain:

```solidity
pragma solidity 0.4.20;

contract Test {
    uint256 public a = 1;

    function some() view external {
        assert(a == 1);
        this.other();
        assert(a == 2);
    }

    function other() external {
        a = 2;
    }
}
```
Unsure how this works on Solidity level anyway.

cd1m0:
> Approve this. Still, there maybe some inconsistent behavior in case of older Solidity (`0.4`) that may allow calling changing state function from non-chaning state function. Effects of this case are uncertain:
> 
> ```solidity
> pragma solidity 0.4.20;
> 
> contract Test {
>     uint256 public a = 1;
> 
>     function some() view external returns(uint256) {
>         assert(a == 1);
>         this.other();
>         assert(a == 2);
>         return a;
>     }
> 
>     function other() external {
>         a = 2;
>     }
> }
> ```
> 
> Unsure how this works on Solidity level anyway. My current idea is that stage is modified in memory, but doesn't get to a storage due to function limitations. In this example calling `some()` would return `2`, BUT you can call `some()` multiple times, therefore `a` never actually stores `2` (until you call `other()` via tx call).

Good point. I think given how outdated Solidity 0.4.x is, we can safely ignore this issue. But good catch, I had forgotten that was not enforced back then

